PR #9, opened 2022-10-08
Update README and add example scripts

sven-oly:


sven-oly:
Submitting without review

PR #10, opened 2022-10-10
Fix rust main executor

sven-oly:
Resolves problem while running under the test driver.


sven-oly:
I'm submitting this without review.

PR #15, opened 2022-10-17
Sync verifier

sven-oly:


executors/python/executor.py
```diff
@@ -7,24 +7,31 @@
...
   def handle_input(self):
     while True:
       inline = sys.stdin.readline()
       if self.debug:
         print('#INPUT = >%s<' % inline)
-      if inline == "#EXIT":
+      if inline[0:5] == "#EXIT":
         break
-      elif inline == '#VERSION':
+      elif inline[0:8] == '#VERSION':
```
gnrunge, line 19: Ditto

executors/python/executor.py
```diff
@@ -7,24 +7,31 @@
...
-    self.debug = 1
+    self.debug = None
 
   def handle_input(self):
     while True:
       inline = sys.stdin.readline()
       if self.debug:
         print('#INPUT = >%s<' % inline)
-      if inline == "#EXIT":
+      if inline[0:5] == "#EXIT":
```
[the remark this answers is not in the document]
sven-oly: It's correct as is.

executors/python/executor.py
sven-oly: Ditto

PR #31, opened 2022-10-26
Moving every nodejs references to node

sven-oly:
This renames all code and files from "nodejs" to "node", making it possible to run Github Actions with that platform. This includes code to create output directories in testResults and testReports as needed.


DDT_DATA/testReports/node/coll_test_shift.json
```diff
@@ -0,0 +1 @@
+{"title": "", "platform": {"platform": "NodeJS", "platformVersion": "v18.7.0", "icuVersion": "71.1"}, "test_environment": {"test_language": "node", "executor": "node ../executors/node/executor.js", "test_type": "coll_shift_short", "datetime": "10/26/2022, 11:07:43", "timestamp": "1666807663.9496472", "inputfile": "../DDT_DATA/testData/coll_test_shift.json", "resultfile": "../DDT_DATA/testResults/node/coll_test_shift.json", "icu_version": "ICUVersion.ICU71", "cldr_version": "CLDRVersion.CLDR41", "test_count": "8"}, "test_errors": [], "timestamp": "2022-10-26 11:07", "failCount": 0, "passCount": 8, "failingTests": [], "missing_verify_data": [], "test_error_count": 0}
```
sffc, line 1: This PR is adding a lot of new files.

testdriver/ddtargs.py
```diff
@@ -2,7 +2,7 @@
 
 # Args expected with examples
 #  test_type: coll_shift, decimal_fmt, etc.
-#  exec: nodejs, rust, cpp, java, custom
```
sffc, line 5: Note: it's okay for these identifiers to differ from the command line arguments. It's also okay to change them. But, the Node.js binary being named "node" does not imply that we need to change the name here.

sffc (approved):

PR #32, opened 2022-10-27
Script to get testdata and run tests. Fix testdriver and verifier files.

sven-oly:
This script runs the test generator, storing testData in a temporary directory. Then it runs testdriver on known tests, creating testResults. These are then verified, producing testReports.

When ready, the test results and test reports will be uploaded to Cloud Storage.

Finally, the temporary directory is removed.


sven-oly:
I'm submitting this, fixing several issues in testdriver and verifier. The script can the be added as another GHA.

PR #33, opened 2022-10-27
Update rust.yml to include cargo build

sven-oly:


sven-oly:
What do you think? Do we need to do `cargo build` in the checkin? And please remind me how we restrict this to only triggering when the Rust executor changes.

sffc:
> What do you think? Do we need to do `cargo build` in the checkin?

I think `cargo clippy` should be sufficient. I think it catches compile errors. If we also wanted to run a build for the purposes of catching compile errors, then `cargo check` is the right command, not `cargo build`

> And please remind me how we restrict this to only triggering when the Rust executor changes.

That configuration is at the top of rust.yml and it applies to all jobs in that file.

sven-oly:
I'm closing this PR because it's not necessary.

PR #35, opened 2022-11-07
Summary: lint Python code, fix report formats, update full generate-test-verify script

sven-oly:


sffc (approved):
I'd like to see the Python lint enforced in CI; there's not much use in linting/formatting code if there's nothing to keep it enforced

PR #53, opened 2023-03-21
Fail if #VERSION check cannot be done

sven-oly:


sven-oly:
Fixing test driver to stop testing a platform if the #VERSION fails, avoiding testing all the cases.

PR #60, opened 2023-05-11
Fix Rust build errors

echeran:
There seems to be a version ambiguity that is affecting the Rust compilation on CI.

PR #59 is [failing CI](https://github.com/unicode-org/conformance/actions/runs/4942064491/jobs/8835849491) in the Rust build even though it has not updated Rust code.

### DIFF
--- a/generateDataAndRun.sh
+++ b/generateDataAndRun.sh
@@ -18,7 +18,7 @@ mkdir -p $TEMP_DIR/testData
 # Compile Rust executor code for ICU4X 1.0
 pushd executors/rust/
 cargo clean
-cargo build --release --profile icu4x_1_0
+rustup run 1.61 cargo build --release
 popd
 
 #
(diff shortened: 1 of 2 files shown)


generateDataAndRun.sh
```diff
@@ -15,9 +15,10 @@ rm -rf $TEMP_DIR
 # Clear out old data, then create new directory and copy test / verify data there
 mkdir -p $TEMP_DIR/testData
 
-# Compile Rust executor code
+# Compile Rust executor code for ICU4X 1.0
 pushd executors/rust/
-cargo build --release
+cargo clean
+cargo build --release --profile icu4x_1_0
```
sffc, line 21: Suggestion: use rustup instead of cargo profiles

sffc (approved):

sven-oly (approved):

PR #65, opened 2023-05-17
Add Dart to executors

mosuem:
Just testing collators on the web for now; as ICU4X integration progresses native tests will be possible.


executors/dart_web/bin/executor.exe
sven-oly, line 1: Should this binary be part of the PR? Or should it be be created on the fly but not saved?

generateDataAndRun.sh
sven-oly, line 1: Old lines 44-46 need to be kept. Just add the new line 61 for running dartweb.

testdriver/testplan.py
sven-oly, line 1: Does the dartweb version need to exit immediately after the #VERSION command?

sven-oly (commented):
A few comments and notes. I don't know enough about Dart to review in more detail.

executors/dart_web/out/executor.js
```diff
@@ -0,0 +1,191 @@
+/* Main execution program for Data Driven Test of ICU / Unicode / CLDR
```
mosuem, line 1: This is just a copy of the node executor, I don't know what the smartest way is to not duplicate the file. Copying at runtime seems ugly...

echeran:
@mosuem The PR looks fine in theory, so if we can get this working, we can merge. Note: I don't know yet if this will update our dashboard HTML templates to include the Dart {Web, Native} test results in the dashboards, but I think that is a good for a followup PR, not here.

I checked out the PR (`gh pr checkout 65`) and ran it locally (`./generateDataAndRun.sh`), and I got the following output:

```
...
Resolving dependencies... 
The current Dart SDK version is 3.0.5.

Because dart requires SDK version ^3.1.0-39.0.dev, version solving failed.
```

Can you help resolve this? 

I don't know if we can count on Dart being installed on the Github Actions CI's [runner VMs](https://github.com/actions/runner-images) by default. Or, at least, I don't currently know how to find out which things get pre-installed.

The safest way is to use the [Github Action by the Dart Lang team](https://medium.com/dartlang/announcing-dart-support-for-github-actions-3d892642104) that does the setup of Dart within a CI workflow.

I think what you need to do is edit `.github/worfklows/e2e.yml`, find the job that runs the command `./generateDataAndRun.sh`, and follow the [Dart Github Action docs](https://github.com/marketplace/actions/setup-dart-sdk) to add a step that looks something like

```
...
      - uses: dart-lang/setup-dart@v1
        with:
          sdk: '3.0.5'
...
```

mosuem:
@echeran I updated the Dart version, and added it to the Github CI pinned by a hash (see https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

executors/dart_web/out/executor.js
sven-oly: OK for now, but copy will be a problem when the node version gets out of sync with Dart implementation.

sven-oly (approved):
With the exception of copying the executor.js file, this looks OK to me.

mosuem:
Any idea on how to better fix this, especially until all `test_type`s have been implemented in Dart? I don't have write access, so I can't merge the PR.

sven-oly:
Hey, please resolve the conflict with testdriver/datasets.py
 and then I can merge this!

sven-oly (approved):

mosuem:
@sven-oly Could you approve the running of workflows?

sven-oly:
Shane or Elango, please help with the workflow. I don't see a way for me to approve as Mortiz requested

sffc:
I ran the CI!

@echeran @sven-oly what are the next steps before we can merge this?

PR #67, opened 2023-05-25
Add charts and pagination

sven-oly:
Lots of changes to support data from versions ICU 70, 71, 72, and 73. Updates to directory structures. Updated Verify Summary page.

Pagination on report details not yet working.


verifier/report_template.py
```diff
@@ -12,8 +12,22 @@ def __init__(self):
     <meta charset="UTF-8">
```
sffc, line 12: Out of scope for now, but it would be better if you save big templates like this in a .html file and then imported them into Python, instead of having a big python string like this. It should help your tooling (e.g., syntax highlighting, auto format, etc).

sven-oly:
I don't know why the old generate script is running with the new PR because generateDataAndRun.sh in the new PR is updated. Let's review the GHAs for this.

testdriver/datasets.py
```diff
@@ -168,13 +168,31 @@ class CppVersion(Enum):
 class ICU4XVersion(Enum):
   ICU4XV1 = "1.0"
 
+# Versions for known executors
+# TODO: combine the version info
+IcuVersionToExecutorMap = {
```
echeran, line 173: praise: I like the consolidation of data/state into fewer places

testdriver/datasets.py
```diff
@@ -168,13 +168,31 @@ class CppVersion(Enum):
...
+    'icu4j': {},
 
+}
 # What versions of NodeJS use specific ICU versions
 # https://nodejs.org/en/download/releases/
 NodeICUVersionMap = {
-    "18": "72.1",
-    "16": "71.1",
-    "14": "70.1",
+    "20.1.0": "73.1",
```
echeran, line 191: suggestion: express derivative data as a transformation via code

With the correct structuring of `IcuVersionToExecutorMap`, now `NodeICUVersionMap` should be derivative. According to [this SO answer](https://stackoverflow.com/questions/483666/reverse-invert-a-dictionary-mapping), we should be able to rewrite this as 

```python
NodeICUVersionMap = {v: k for k, v in IcuVersionToExecutorMap['node'].items()}
```

testdriver/testplan.py
```diff
@@ -50,6 +53,10 @@ def __init__(self, exec_data, test_type, args=None):
 
   def setOptions(self, options):
     self.options = options
+    try:
+      self.icu_version = options.icu_version
+    except:
+      logging.warn('NO ICU VERSION SET')
```
echeran, line 59: praise: the incremental improvements along the way are nice

testdriver/testplan.py
```diff
@@ -71,21 +78,36 @@ def runPlan(self):
...
+    # If icu_version is "latest" or not set, get the highest numbered
+    # version of the test data
+    input_root = os.path.join(self.options.file_base,
+                              self.options.input_path)
+    icu_test_dirs = glob.glob('icu*', root_dir=input_root)
+    if not icu_test_dirs:
+      raise Exception('No ICU test data found in directory %s' % input_root)
+
+    if self.icu_version not in icu_test_dirs:
+      newest_version = sorted(icu_test_dirs, reverse=True)[0]
```
echeran, line 90: Are we okay to assume that the string representation of the version number can be sorted without changing the correct logical order?

FYI, according to [this SO answer](https://stackoverflow.com/a/2574230), there's a built-in library to sort strings of semantic versions.

testdriver/datasets.py
sven-oly: thx

testdriver/testplan.py
sven-oly: Ok

testdriver/testplan.py
sven-oly: Good idea to fix this. However, for now the test dirs are labeled icu69, icu70, icu71, so this should be documented. Adding to the code for next push.

verifier/verifier.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
echeran, line 1: please remove this IDE settings file

verifier/verifier.py
```diff
@@ -29,6 +29,9 @@ def __init__(self):
     # All the items that will be verified
     self.verify_plans = []
 
+    # Filename used for the json version of verifier output
+    self.report_filename = 'verifier_test_report.json'
```
echeran, line 33: suggestion: convert this from a field to a global constant b/c I don't think you expect this to change from instance to the next at runtime

PR #77, opened 2023-06-23
Add languageNameTable for icu67 to icu70

sven-oly:
Obtained from Google's processing of CLDR release data. This supports tests of lang_names.


echeran:
Needs review regarding diffs w/ public CLDR to inform suitability of use. @macchiati @markusicu

sven-oly:
I've reviewed the generator for this data file. There is no Google-specific data added or changed as far as I can see. I think this is OK to make public.

The one except is that some alternative names may be used.

PR #80, opened 2023-06-29
Fix node numberformat on node

sven-oly:
This resolves most of the test failures with node and numberformat in ICU73. The number of unsupported options is large.


executors/node/numberformat.js
```diff
@@ -80,36 +145,75 @@ module.exports = {
...
     } else {
-      // Check each option for implementation.
+      // Set default maximumFractionDigits with these exceptions
+      if (options['notation'] && options['notation'] == 'compact' ||
+          options['style'] && options['style'] == 'currency' ||
+          options['maximumFractionDigits']) {
+        // Don't set maximumFractionDigits
+      } else {
+        options['maximumFractionDigits'] = 6;  // Default
+      }
```
sffc, line 155: Adding `maximumFractionDigits: 6` should be in test generation, not in the executor

executors/node/numberformat.js
```diff
@@ -80,36 +145,75 @@ module.exports = {
...
+      // ?????
+      if (options['roundingMode'] == undefined) {
+        options['roundingMode'] = 'halfTrunc';
+      }
```
sffc, line 159: Also should probably be in test generation. Could go either way though: do we want to define the "ambient" behavior of roundingMode and maximumFractionDigits for all numberformat tests, or do we just want to specify those behaviors explicitly? I lean toward making them explicit. If they _do_ have defaults, they should adopt the ECMA-402 defaults.

sven-oly:
Please take a look. I'd love to get this PR finished!

echeran (approved):
LGTM

PR #81, opened 2023-07-07
ICU4X: compact decimal formatting, lang display names

sven-oly:
Working toward extending ICU4X number formattting with compact decimal.

Also adding language display names - not working yet!

### DIFF
--- a/executors/rust/Cargo.toml
+++ b/executors/rust/Cargo.toml
@@ -12,15 +12,29 @@ env_logger = "0.9.1"
 
 substring = "1.0"
 json = "0.12.4"
-serde = { version = "1.0.143", features = ["derive"] }
+serde = {version = "1.0.143", features = ["derive", "alloc"], optional=true}
 serde_json = "1.0.48"
 rustc_version_runtime = "0.1.*"
     
-fixed_decimal = {version="0.5.0"}
-writeable =  {version="0.5.0"}
-
-icu = { version="1.0.0", features = ["serde"] }
-icu_testdata = { version="1.0.0", features = ["metadata"] }
-icu_provider = { version="1.0.0"}
-
+fixed_decimal = {version="0.5.3"}
+writeable = {version="0.5.2"}
+icu_compactdecimal = {version="0.2.0"}
+icu_displaynames = {version="0.10.0"}
+    
+icu = {version="1.2.0", features = ["serde"] }
+icu_testdata = {version="1.2.0"}
+icu_compactdecimal_data = {git="https://github.com/unicode-org/icu4x", rev="5fcb6c19f72b2e5ad1d49b1a2332774c6d68c25d", optional=true}
+icu_provider = {version="1.2.0", features=["macros"]}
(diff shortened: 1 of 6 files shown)


executors/rust/Cargo.toml
```diff
@@ -12,15 +12,29 @@ env_logger = "0.9.1"
...
+icu_decimal = { version = "1.2.0", optional=true}
+databake = { version = "0.1.3", features = ["derive"], optional = true}
+zerovec = { version = "0.9.4", features = ["yoke"] }
+    
 [features]
+default = ["compiled_data"]
+std = ["fixed_decimal/std", "icu_decimal/std", "icu_plurals/std", "icu_provider/std"]
+serde = ["dep:serde", "zerovec/serde", "icu_decimal/serde", "icu_plurals/serde"]
+datagen = ["std", "serde", "dep:databake", "zerovec/databake"]
+compiled_data = ["dep:icu_compactdecimal_data", "dep:icu_locid_transform", "icu_decimal/compiled_data", "icu_plurals/compiled_data"]
```
sffc, line 40: Issue: As a client crate you shouldn't need features. It looks like you copied these from an ICU4X crate and they don't make sense here.

executors/rust/Cargo.toml
```diff
@@ -12,15 +12,29 @@ env_logger = "0.9.1"
...
-icu_testdata = { version="1.0.0", features = ["metadata"] }
-icu_provider = { version="1.0.0"}
-
+fixed_decimal = {version="0.5.3"}
+writeable = {version="0.5.2"}
+icu_compactdecimal = {version="0.2.0"}
+icu_displaynames = {version="0.10.0"}
+    
+icu = {version="1.2.0", features = ["serde"] }
+icu_testdata = {version="1.2.0"}
```
sffc, line 25: Issue: You should only need to depend on `icu@1.2`, `icu_testdata@1.2`, and maybe `fixed_decimal`, `writeable`, and `icu_locid_adapters` if you need to reference their types.

executors/rust/src/numberfmt.rs
```diff
@@ -71,14 +85,34 @@ pub fn run_numberformat_test(json_obj: &Value) -> Result<Value, String> {
...
+            CompactDecimalFormatter::try_new_long_unstable(
+                &provider,
+                &data_locale,
+                Default::default(),
+            ).unwrap()  // ??  options);
+        }
+        cdf.format_i64(input_num)
+    }
+    else {
+        let input_num = input.parse::<FixedDecimal>().map_err(|e| e.to_string())?;
```
sffc, line 110: Note that we will need to set the minimum/maximum significant/fraction digits somewhere around here.

sffc (changes requested):

sffc (commented):
The ICU4X integration is looking good; next step is to tweak exactly how we generate options for the ICU4X APIs and label test cases as unsupported.

sffc (approved):

sffc (approved):

PR #87, opened 2023-08-10
Fix dart

mosuem:
- Add version string of the `intl4x` dependency by parsing the `pubspec.lock` file
- Add locale to avoid having `dart_web` looking for the users locale in the `window` object, which doesn't exist in NodeJS.


sven-oly (approved):
Thanks for these updates.

PR #88, opened 2023-08-10
Fix verifier to show dart_web output

sven-oly:
Updates summary report to correctly show results from dart_web. This now shows graphs and results for all three platforms!

### DIFF
--- a/generateDataAndRun.sh
+++ b/generateDataAndRun.sh
@@ -41,10 +41,10 @@ rustup install 1.61
 rustup run 1.61 cargo build --release
 popd
 
-pushd executors/dart_native/
-dart pub get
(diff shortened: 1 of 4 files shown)


sven-oly:
PTAL to make the DART results appear publicly!

generateDataAndRun.sh
```diff
@@ -41,10 +41,10 @@ rustup install 1.61
 rustup run 1.61 cargo build --release
 popd
 
-pushd executors/dart_native/
-dart pub get
-dart compile exe bin/executor.dart
-popd
+#pushd executors/dart_native/
```
sffc, line 44: comment?

sffc (approved):

PR #93, opened 2023-08-18
Add likely_subtags: data generation, node executor

sven-oly:


sven-oly:
This uses the latest CLDR data in https://github.com/unicode-org/cldr/blob/main/common/testData/localeIdentifiers/likelySubtags.txt for checking Node's minimize and maximize locale tags.

testgen/testdata_gen.py
```diff
@@ -172,6 +172,83 @@ def generateLanguageNameTestDataObjects(self, rawtestdata, json_tests, json_veri
...
+            # split at ";" and ignore whitespace
+            tags = list(map(str.strip, line.split(';')))
+            # This should give 3 items
+            if len(tags) < 3:
+                logging.warning('Likely subtags: count of tages != 3: %s',
+                                line)
+                continue
+
+            # Create 3 minimize tests
+            expected = tags[2]
```
sffc, line 211: Suggestion: you should read `tags[3]` and put it in an option called `minimizeFavorScript`

testgen/testdata_gen.py
```diff
@@ -172,6 +172,83 @@ def generateLanguageNameTestDataObjects(self, rawtestdata, json_tests, json_veri
...
+            for index in range(0, 3):
+                label = str(count).rjust(max_digits, '0')
+                locale = tags[index]
+                test_min = {'label': label,
+                         'locale': locale,
+                         'option': 'minimize'
+                         }
+                verify = {'label': label,
+                         'verify': expected
+                          }
```
sffc, line 221: Please note the highlighted portions of the notice at the top of the test data file:

> Source ;	AddLikely ;	RemoveFavorScript ;	RemoveFavorRegion
>   Source: a locale to which the following operations are applied.
>   AddLikely: the result of the Add Likely Subtags.
>                      If Add Likely Subtags fails, then “FAIL”.
>   RemoveFavorScript: Remove Likely Subtags, when the script is favored.
>                      ***Only included when different than AddLikely.***
>   RemoveFavorRegion: Remove Likely Subtags, when the region is favored.
>                      ***Only included when different than RemoveFavorScript.***

I don't immediately see where you are doing the "only included when..." portion of the instructions. Most but not all cases have data populated for RemoveFavorScript, and only a few have data populated for RemoveFavorRegion

testgen/testdata_gen.py
sven-oly: Thanks. Will do!

testgen/testdata_gen.py
sven-oly: Updated tests and code in latest commits.

executors/node/likely_subtags.js
```diff
@@ -0,0 +1,43 @@
...
+
+    try {
+      let result_locale;
+      if (test_option === 'maximize') {
+        result_locale = intl_locale.maximize().baseName;
+      } else if (test_option === 'minimizeFavorScript' ||
+                 test_option === 'minimize') {
+        result_locale = intl_locale.minimize().baseName;
+      } else if (test_option === 'minimizeFavorRegion') {
+        result_locale = intl_locale.minimizeFavorRegion().baseName;
```
sffc, line 33: Note: there is no `Intl.Locale.prototype.minizeFavorRegion`; this should be unsupported

testgen/testdata_gen.py
```diff
@@ -172,6 +172,103 @@ def generateLanguageNameTestDataObjects(self, rawtestdata, json_tests, json_veri
...
+            tags = list(map(str.strip, line.split(';')))
+            # This should give 3 items
+            if line.find('FAIL') >= 0:
+                # What do do with this?
+                pass
+
+            # Create minimize tests - default is RemoveFavorScript
+            if tags[2]:
+                remove_favor_script = tags[2]
+                for tag in tags:
```
sffc, line 212: This generates extra cases for each line. Probably not needed.

testgen/testdata_gen.py
sven-oly: Updated to handle non-included values, falling back to previous.

testgen/testdata_gen.py
sven-oly: Done!

testgen/testdata_gen.py
```diff
@@ -172,6 +172,103 @@ def generateLanguageNameTestDataObjects(self, rawtestdata, json_tests, json_veri
...
+            # Ignore blank and # comment lineslines()
+            if len(line) == 0 or line[0] == "#":
+                continue
+            # split at ";" and ignore whitespace
+            tags = list(map(str.strip, line.split(';')))
+            # This should give 3 items
+            if line.find('FAIL') >= 0:
+                # What do do with this?
+                pass
+
```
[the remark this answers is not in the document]
sven-oly: Removed this, using "FAIL" as if it's expected data.

executors/node/likely_subtags.js
sven-oly: Updated now.

testgen/testdata_gen.py
sven-oly: Simplified, creating maximum of 3 cases per source locale.

testgen/testdata_gen.py
```diff
@@ -172,6 +172,103 @@ def generateLanguageNameTestDataObjects(self, rawtestdata, json_tests, json_veri
...
+            # Create minimize tests - default is RemoveFavorScript
+            if tags[2]:
```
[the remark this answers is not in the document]
sven-oly: Done

sven-oly:
Much better now! PTAL

executors/node/likely_subtags.js
sffc: I don't see where this is fixed but OK to fix later

sffc (approved):

PR #94, opened 2023-08-18
Adding coll_nonignorable_short testing

sven-oly:
This adds collation for node, comparing with default parameters. There are still test failures that need to be explained.

And rust doesn't yet implement this option. Note that this could be redone using a runtime parameter to the collation executor rather than separate collation .js code.

### DIFF
--- a/executors/node/collator_nonignorable.js
+++ b/executors/node/collator_nonignorable.js
@@ -1,0 +1,44 @@
+// The Collator for nonignorable short tests.
+
+// !!! TODO: Collation: determine the sensitivity that corresponds
+// to the strength.
+module.exports = {
+
+  testCollationNonignorableShort: function(json) {
+
+    // Global default locale
+    let testLocale = '';
+    let testCollOptions = {};
+
+    // Set up collator object with optional locale and testOptions.
+    let coll;
+    try {
+      if (testLocale) {
+        coll = new Intl.Collator(testLocale, testCollOptions);
+      } else {
+        coll = new Intl.Collator(testCollOptions);
+      }
+      let d1 = json['string1'];
+      let d2 = json['string2'];
+
+      const compared = coll.compare(d1, d2);
+      let result = compared<= 0 ? true : false;
+      let resultString = result ? true : false;
+      outputLine = {'label':json['label'],
+                    "result": result
+                   }
+
+      if (result != true) {
+        // Additional info for the comparison
+        outputLine['compare'] = compared;
+      }
+
+    } catch (error) {
+      outputLine =  {'label': json['label'],
+                     'error_message': error.message,
+                     'error': 'Collator compare failed'
+                 };
+    }
+    return outputLine;
+  }
+};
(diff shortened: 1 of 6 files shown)


executors/node/collator_nonignorable.js
```diff
@@ -0,0 +1,44 @@
...
+    let testCollOptions = {};
+
+    // Set up collator object with optional locale and testOptions.
+    let coll;
+    try {
+      if (testLocale) {
+        coll = new Intl.Collator(testLocale, testCollOptions);
+      } else {
+        coll = new Intl.Collator(testCollOptions);
+      }
```
sffc, line 20: The collator needs to be created with locale `undefined`; it does not automatically detect that the first argument is not a locale

executors/node/collator_nonignorable.js
```diff
@@ -0,0 +1,44 @@
+// The Collator for nonignorable short tests.
```
sffc, line 1: Question: can you explain why you made this a new executor instead of adding it to the existing coll_shift_short executor? It would be nice to make clearer when to add things to an existing executor versus creating a new executor

sven-oly: Yup, I've refactored this to use the same collator.js functions with a parameter. Same for ICU4X version.

sffc: I think you should keep a single collator testType.

executors/dart_web/bin/collator.dart
```diff
@@ -22,10 +22,20 @@ String testCollationShort(String jsonEncoded) {
     } else {
       coll = Intl();
     }
-    var d1 = json['string1'];
-    var d2 = json['string2'];
+    var d1 = json['s1'];
```
mosuem, line 25: This should work:
```
var d1 = json['s1'];
var d2 = json['s2'];

var collationOptions = CollationOptions(
      ignorePunctuation: json['ignorePunctuation'] ?? false,
);
```

mosuem: (If the json contains a string and not a bool, it should rather be
```
var d1 = json['s1'];
var d2 = json['s2'];

var collationOptions = CollationOptions(
  ignorePunctuation: bool.tryParse(json['ignorePunctuation']) ?? false,
);
```
)

sven-oly: Thanks. I'm integrating these, changing "var" to "final".

sven-oly: This now appears to work with dart_web. Thank you!

sven-oly:
Ready for another review. This now works in dart_web, node, and icu4x (mostly) for over 380K test samples.

mosuem (approved):

sven-oly:
OK, folks. Please take a look. This has a lot of changes in the testing data for collation. The quick tests have been simplified and also are brought up to work with ICU73 data.

sven-oly:
Going ahead to merge this.

PR #97, opened 2023-08-23
Make `generateDataAndRun.sh` configurable

mosuem:
Accept a json file, either `run_config.json` as a default or the first arg passed, such as `generateDataAndRun.sh other_json_file.json`.


sven-oly:
Thanks for working on this. It's useful and makes the generator script much cleaner.

sven-oly (approved):

PR #99, opened 2023-09-12
icu4X likely subtags added

sven-oly:
Thanks, Elango, for helping me get this to work!


executors/rust/src/likelysubtags.rs
```diff
@@ -0,0 +1,41 @@
...
+    let locale_str: &str = json_obj["locale"].as_str().unwrap();
+    
+    let mut locale = Locale::from_str(locale_str).unwrap();
+
+    if test_option == &"minimize" {
+        lc.minimize(&mut locale);
+    } else
+        if test_option == &"maximize" {
+            lc.maximize(&mut locale);
+        }
```
sffc, line 33: Should we have an `else` block for a `test_option` that we don't understand? (like minimizeFavorRegion or minimizeFavorScript)

PR #113, opened 2023-10-17
Rust number format: mark scientific as unsupported

sven-oly:


sven-oly:
This will move over 600 failures to "unsupported".

sffc (approved):

PR #114, opened 2023-10-18
Rust likely_subtags: categorize"minimizeFavorRegion" as "unsupported"

sven-oly:


sven-oly:
This updates the schema for results, adding error fields in test output.

sven-oly:
This will move about 800 ICU4X failing tests to unsupported for likely subtags, leaving about 200.

sffc (commented):
I believe minimizeFavorRegion is unsupported in _all_ current executors so we should mark it as such in the rest of them, too.

echeran (approved):
LGTM

PR #119, opened 2023-11-08
Clean up Rust executor

robertbastian:


robertbastian:
@sven-oly

sven-oly:
Thanks, Rob, for your suggested changes! Shane, please review also!

executors/rust/src/decimalfmt.rs
```diff
@@ -35,17 +30,17 @@ pub fn run_numberformat_test(json_obj: &Value) {
     options.grouping_strategy = options::GroupingStrategy::Min2;
 
     let fdf = FixedDecimalFormatter::try_new_with_buffer_provider(
-        &provider, &data_locale, options)
+        &provider, &langid.into(), options)
```
echeran, line 33: I prefer using the previous `DestType::from(x)` rather than `x.into()` because it's explicit to the reader about which type you get back.

executors/rust/src/displaynames.rs
```diff
@@ -24,14 +21,13 @@ pub fn run_coll_test(json_obj: &Value) {
...
     };
-    let data_locale = DataLocale::from(langid);
 
     let data_provider = icu_testdata::unstable();
 
     let mut options = DisplayNamesOptions::new();
 
     let displaynames: DisplayNames =
-        DisplayNames::try_new_unstable(&data_provider, data_locale, options).unwrap();
+        DisplayNames::try_new_unstable(&data_provider, langid.into(), options).unwrap();
```
echeran, line 30: ditto, here & throughout

echeran (commented):
Most changes look really good. My only comment is about preferring `From` over `Into` on the basis of explicit > implicit.

executors/rust/src/decimalfmt.rs
robertbastian: I agree in the general case, I find `.into()`s confusing. However in the context of `DataLocale` in ICU4X, we usually (always?) use `.into()` in our docs, because `DataLocale` is not meant to be a super public type. It required importing the `icu_provider` crates, which for most users (such as here), is an implementation detail.

echeran: Okay, in that case, LGTM

echeran (approved):

PR #124, opened 2023-11-17
Advanced collation - fixes many NodeJS collation failures

sven-oly:


verifier/testreport.py
```diff
@@ -572,6 +572,15 @@ def characterize_failures_by_options(self, failing_tests):
...
 
+                for key in ['language_label', 'ignorePunctuation', 'compare_result', 'compare_type', 'test_description']:
+                    if test.get(key):  # For collation results
+                        value = test[key]
+                        if key not in results:
+                            results[key] = {}
+                        if value in results[key]:
+                            results[key][value].append(label)
+                        else:
+                            results[key][value] = [label]
```
echeran, line 583: optional: if you initialize `results` using [Python's support for auto-initializing the values in a nested dictionary](https://stackoverflow.com/questions/651794/whats-the-best-way-to-initialize-a-dict-of-dicts-in-python), such as:

```
results = collections.defaultdict(lambda : collections.defaultdict(list))
```

...then the body of this block of code would become a single line.

```suggestion
                for key in ['language_label', 'ignorePunctuation', 'compare_result', 'compare_type', 'test_description']:
                    results[key][value].append(label)
```

echeran (approved):

verifier/testreport.py
sven-oly: Thx. Will resolve later.

PR #125, opened 2023-11-17
Replace `python` with `python3`

echeran:
...according to the common wisdom that it is preferable to explicitly use `python3` when invoking Python 3, rather than just `python`.


sven-oly (approved):
Yay!

PR #134, opened 2023-11-22
Add ICU4J executor skeleton implementation

echeran:
Add an empty implementation for an ICU4J executor.  Also, add an option to running an executor (called `--ignore`) that allows a new executor instance to be skipped.


sven-oly (approved):
Thanks for getting started with the Java!

PR #136, opened 2023-11-23
Reduce standard output printing

echeran:
Reduce printing to standard output by:

- Adjusting logging levels appropriately
- Converting plain `print()` statements to logging, using appropriate level
- Change log rotation strategy
    - Our Python logging config has more verbose (DEBUG level) logging going to file than the concise (INFO level) logging subset allowed to be repeated to standard output
    - We can rotate the log files so that we can keep a short term history of the output from previous runs
    - The Python `logging.config` may have been invoked multiple times in one end-to-end run (ex: once per executor invocation), because we're not running a single Python executable, but instead multiple instances of smaller disjoint Python executables. Simpler system, but flexibility trades off with convenience. The consequence is log files might be getting overwritten in undesirable unexpected ways
    - Drop the log rotation behavior from the Python logger
    - Use the Linux/Unix tool `logrotate` in a manual way, once per program execution

Overall result: stdout drops from [~30K lines](https://github.com/unicode-org/conformance/actions/runs/6962036710/job/18944907211?pr=134) to only ~650 lines. Full details are still captured across all of the log files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,4 +31,6 @@ testgen/*.json
 
 TEMP_DATA/*
 
+# log files (including logrotate state and file backups)
 debug.log*
+logrotate.state
--- a/README.md
+++ b/README.md
@@ -252,6 +252,10 @@ Requirements to run Data Driven Testing code locally:
         ```
         rustup update
         ```
+  - Install `logrotate`
+      ```
+      sudo apt-get install logrotate
+      ```
 
 # History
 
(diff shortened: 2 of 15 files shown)


echeran:
Note: A couple of cool things about looking at the log files in VS Code:

- VS Code automatically refreshes the file buffer as new contents get appended to the file. 
- VS Code does automatic color highlighting of `[INFO]`, `[WARNING]`, `[ERROR]`, etc. 

It's like all the benefit of `tail -f` but the benefits of an IDE.

echeran:
If this change looks good, then once it gets merged, feel free to clean up the previous log files by running
```
find -type f -name "debug*" -exec rm {} \;
```

sven-oly (approved):

PR #142, opened 2023-12-06
Fixing charactersizations for collation

sven-oly:


sven-oly:
I need to have these characterizations to better debug CPP additions to Conformance testing.

PR #143, opened 2023-12-08
Updating collation generation and execution

sven-oly:
This updates some of the tests to reset the base string to empty when a new comparison set is started.

This improves passing tests in
Node icu74: from 372,031 --> 372,129
Dart icu73:  from 371,961 --> 372,062
No change in ICU4X, however.

A small improvement, removing about 5% of the failing tests.

### DIFF
--- a/testgen/testdata_gen.py
+++ b/testgen/testdata_gen.py
@@ -713,7 +713,7 @@ def generateCollTestData2(filename,
         is_compare = compare_pattern.match(line_in)
         compare_type = None
         if is_compare:
-            # Reset string1 to the empty string.
+            # Initialize string1 to the empty string.
             string1 = ''
             compare_mode = True
             info = is_compare.group(1)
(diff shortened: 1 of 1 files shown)


testgen/testdata_gen.py
```diff
@@ -714,7 +711,10 @@ def generateCollTestData2(filename,
             continue
 
         is_compare = compare_pattern.match(line_in)
+        compare_type = None
         if is_compare:
+            # Reset string1 to the empty string.
```
echeran, line 716:
```suggestion
            # Initialize string1 to the empty string.
```

echeran: Looks like this is the initialization before we begin the while loop, and we only update `string1` at the end of the loop body.

echeran (approved):
LGTM. apply the nitpick about the comment wording if you want

PR #144, opened 2023-12-11
Cleaning up Rust executor

robertbastian:
Removed unnecessary deps, included dead modules, moved version detection into build.rs

### DIFF
--- a/executors/rust/1.3/src/main.rs
+++ b/executors/rust/1.3/src/main.rs
@@ -17,22 +17,21 @@
 // https://unicode-org.github.io/icu4x-docs/doc/icu_collator/index.html
 
 mod collator;
+mod decimalfmt;
+mod displaynames;
 mod langnames;
 mod likelysubtags;
 mod numberfmt;
 
-use serde_json::{json, Value};
-
-use std::collections::HashMap;
-
-use std::env;
-use std::io::{self};
-
-// Test modules for each type
 use collator::run_collation_test;
 use langnames::run_language_name_test;
 use likelysubtags::run_likelysubtags_test;
 use numberfmt::run_numberformat_test;
+
+use serde_json::{json, Value};
+use std::collections::HashMap;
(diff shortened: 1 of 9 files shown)


robertbastian:
In 1.5 we can remove the datagen dependency and the build script will be 
```rust
fn main() {
    let metadata = cargo_metadata::MetadataCommand::new().exec().unwrap();
    for package in metadata.packages.iter() {
        if package.name == "icu" {
            println!(
                "cargo:rustc-env=CONFORMANCE_ICU4X_VERSION={}",
                package.version
            );
        } else if package.name == "icu_calendar_data" {
            println!(
                "cargo:rustc-env=CONFORMANCE_ICU_VERSION={}",
                &package.metadata["sources"]["icuexport"]["tagged"]
                    .as_str()
                    .unwrap()
            );
            println!(
                "cargo:rustc-env=CONFORMANCE_CLDR_VERSION={}",
                &package.metadata["sources"]["cldr"]["tagged"]
                    .as_str()
                    .unwrap()
            );
        }
    }
}

```

robertbastian:
@sffc

sffc (approved):

executors/rust/1.3/src/main.rs
```diff
@@ -17,23 +17,22 @@
 // https://unicode-org.github.io/icu4x-docs/doc/icu_collator/index.html
 
 mod collator;
+mod decimalfmt;
+mod displaynames;
```
sffc, line 21: @sven-oly Are these files being used? I think they might be dead code because we have other modules, `numberfmt` and `langnames`, that do these tests.

robertbastian:
I cannot merge

executors/rust/1.3/src/main.rs
sven-oly: If they are dead code, please remove. Thanks, Rob, for this work!

PR #145, opened 2023-12-14
ICU4C lang names working now...

sven-oly:


executors/cpp/main.cpp
```diff
@@ -109,11 +111,19 @@ int main(int argc, const char** argv)
       // else if (test_type == "number_fmt") {
       //   outputLine = test_numfmt(json_input);
       //   // }
-      // else if (test_type == "language_display_name") {
-      //   outputLine = test_langnames(json_input);
-      // }
-else {
-        std::string outputLine = "# BAD TEST " + test_type;
+      else if (test_type == "lang_names" ||
```
echeran, line 114: We should really have only one value in this key (b/c it behaves like an enum) to represent this one component type.

testdriver/testplan.py
```diff
@@ -523,15 +523,17 @@ def send_one_line(self, input_line):
                 # TODO!!!! Return an error for the offending line instead of failing for the whole batch
 
                 return None
-                input = json.loads(input_line.replace('#EXIT', ''))
-                error_result = {'label': input['label'],
-                                'input_data': input,
-                                'error': self.run_error_message
-                }
+                try:
```
echeran, line 526: insert an empty to visually separate the try/except block

echeran: also: do you even need the try/except block here? see if you can remove it

echeran (changes requested):

executors/cpp/main.cpp
sven-oly: Agreed.

executors/cpp/main.cpp
```diff
@@ -41,15 +41,18 @@ using std::string;
 // Test functions
 extern const string test_collator(json_object *json_in);
 // extern const string test_numfmt(json_object *json_in);
-// extern const string test_langnames(json_object *json_in);
+extern const string test_langnames(json_object *json_in);
+
+// !!! extern const string test_likely_subtags(json_object *json_in);
```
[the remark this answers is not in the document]
sven-oly: Done

executors/cpp/langnames.cpp
```diff
@@ -29,38 +29,47 @@ using std::string;
...
-  json_object_object_add(return_json,
-                         "result",
-                         json_object_new_string(name.c_str()));
- string return_str = json_object_to_json_string(return_json);
 
+  char test_result_string[1000] = "";
+
+  int32_t chars_out = testLang.extract(test_result_string, 1000, nullptr, status);
+  if (U_FAILURE(status)) {
+    json_object *error_msg = json_object_new_object();
```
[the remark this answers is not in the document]
sven-oly: Done

testdriver/testplan.py
sven-oly: Changed this. PTAL.

testdriver/testplan.py
```diff
@@ -520,18 +520,15 @@ def send_one_line(self, input_line):
...
+                error_result = {'label': 'UNKNOWN',
+                                'input_data': input_line,
                                 'error': self.run_error_message
-                }
+                                }
                 return json.dumps(error_result)
         except BaseException as err:
             logging.error('!!! send_one_line fails: input => %s<. Err = %s', input_line, err)
-            input = json.loads(input_line.replace('#EXIT', ''))
+            input = json.loads(input_line.replace('#EXIT', '').strip())
```
echeran, line 531: This seems harmless, but why do we need to replace `#EXIT` from the input string when recording the error? In other words, why aren't we faithfully documenting the input line as it is? I wouldn't expect an `#EXIT` line to be the cause of errors, but even if it is, then our logs should be a faithful source of the evidence.

echeran (approved):
Looks fine, LGTM. But take a look at the question about why we're logging a transformed version of the input during an error scenario (and not the original input itself).

testdriver/testplan.py
sven-oly: Let's review this later. Tnx for for the LGTM.

PR #146, opened 2023-12-18
Fix collation testgen for unpaired surrogates

sven-oly:


sven-oly:
This needs to be done to fix problems with some platforms in collation tests.

PR #152, opened 2023-12-20
CPP: Number format added; collation updated

sven-oly:
Preliminary implementation of CPP number formatting - lots to do yet.

Collation testing is improved by about 50%. Now using C++ API instead of C-language.


sven-oly:
I'm going to add this now without review.

Issue #129, opened 2023-11-18
ICU4X Collation failures


sven-oly:
ICU4X in conformance testing shows more that 20% of the tests failing, seen here:
[ICU4X/icu73](https://unicode-org.github.io/conformance/rust/icu73/collation_short/verifier_test_report.html)

The actual collator options are seen in the test failure detail, with a few examples here. The inputs are s1 and s2 and the actual options used are given

- {"label":"0010001","s1":"𑜿!","s2":"𑜿?","line":8661,"ignorePunctuation":true}	CollatorOptions { strength: Some(Tertiary), alternate_handling: Some(Shifted), case_first: None, max_variable: None, case_level: None, numeric: None, backward_second_level: None }
- {"label":"0243300","s1":"𑛁b","s2":"𑜱b","line":47434}	CollatorOptions { strength: Some(Tertiary), alternate_handling: None, case_first: None, max_variable: None, case_level: None, numeric: None, backward_second_level: None }
- {"label":"0373766","s1":"龜a","s2":"龜a","line":177900}	CollatorOptions { strength: Some(Tertiary), alternate_handling: None, case_first: None, max_variable: None, case_level: None, numeric: None, backward_second_level: None }

We need some help debugging help with this!

sffc:
Hey @hsivonen, can you take a look at these? Here is how we configure the ICU4X collator:

```rust
    let mut options = CollatorOptions::new();
    options.strength = Some(Strength::Tertiary);

    // Ignore punctuation only if using shifted test.
    if let Some(ip) = ignore_punctuation {
        if *ip {
            options.alternate_handling = Some(AlternateHandling::Shifted);
        }
    }

    let collator: Collator = Collator::try_new(&locale!("en").into(), options).unwrap();

    let comparison = collator.compare(str1, str2);
```

https://github.com/unicode-org/conformance/blob/34f5f0cee5cacc1d438a05833459feabd5cdf767/executors/rust/src/collator.rs#L17

Here is the corresponding JS code which passes the tests:

```javascript

    // Locale if provided in the test data.
    let testLocale = undefined;
    if ('locale' in json) {
      testLocale = json['locale'];
    }
    let testCollOptions = {};
    if ('ignorePunctuation' in json) {
      testCollOptions = {
        ignorePunctuation:json['ignorePunctuation']}
    }

    // Get other fields if provided
    let rules = undefined;
    if ('rules' in json) {
      rules = json['rules'];
    }

[...]

PR #153, opened 2023-12-22
Rust executor: check is_le instead of is_lt

sffc:
See #129

I made this change directly in the GitHub UI so I haven't tested it yet. Hopefully CI is helpful :)


sven-oly (approved):

sven-oly:
Thank you, Shane.

PR #154, opened 2024-01-04
ICU4C Updating Number Format

sven-oly:


executors/cpp/number_fmt.cpp
```diff
@@ -48,38 +52,248 @@ using std::endl;
...
+
+  string notation_string = "";
+  string precision_string = "";
+  string unit_string = "";
+  string unitDisplay_string = "";
+  string style_string = "";
+  string currency_string = "";
+  string roundingMode_string = "";
+  string compactDisplay_string = "";
+  string conformahceScale_string = "";
```
sffc, line 143: conformahceScale_string?

executors/cpp/number_fmt.cpp
```diff
@@ -91,35 +305,59 @@ const string test_numfmt(json_object *json_in) {
...
+  else {
+    // Use settings to initialize the formatter
+    nf = NumberFormatter::withLocale(displayLocale)
+         .notation(notation_setting)
+         .precision(precision_setting)
+         .integerWidth(integerWidth_setting)
+         .scale(scale_setting)
+         .unit(unit_setting)
+         .unitWidth(unit_width_setting);
+  }
```
sffc, line 334: To generalize, put the common stuff outside the if statement and then do `nf = nf.unit(...)` inside the if

executors/cpp/number_fmt.cpp
```diff
@@ -91,35 +305,59 @@ const string test_numfmt(json_object *json_in) {
...
 
-  // TEMPORARY - just return the input.
-  test_result = input_string;
+  LocalizedNumberFormatter nf;
+  if (notation_string == "scientific") {
+    notation_setting = Notation::scientific();
+  }
 
-  // TODO!!!
-  //  SimpleNumberFormatter snf = SimpleNumberFormatter::forLocale(displayLocale, status);
```
sffc, line 100: This approach is fine but I would also have been okay with an approach that constructs a NumberFormatter from the skeleton string. If we don't use the skeleton string, I think we should remove it from the input data and keep it, at best, in some sort of annotation.

executors/cpp/number_fmt.cpp
sven-oly: I've revised this to merge the two styles.

executors/cpp/number_fmt.cpp
sven-oly: Now working on using the skeleton where it exists.

echeran (approved):
rslgtm

@sven-oly is going to follow up with another PR to address the remaining errors. I'll take a closer look at the code then. But this PR that brings it down from 75% errors to 20% errors is a good first step.

PR #157, opened 2024-01-09
Update ICU number format to use skeletons when available

sven-oly:
Skeletons are available for most data. However, there are some patterns that are not yet converted to skeletons in the data generation.

This PR allows about 97% of the CPP number format tests to pass.
The conversion to options is not completely correct yet for number format.


sven-oly:
Also fix some problems with characterizing results, removing duplication in each result.

echeran (approved):
LGTM

PR #162, opened 2024-01-16
Parallelize test execution and schema checks using Python multiprocessing

sven-oly:


sven-oly:
This reduces the time to run the end to end script. I will add parallel processing to data generation and also verification steps.

schema/schema_validator.py
```diff
@@ -36,74 +48,160 @@ def __init__(self):
...
+            'err_info': None,
+            'test_schema': schema_verify_file,
+            'data_file_name': test_file_name
+        }
+        result = self.validate_json_file(schema_info)
+
+        if isinstance(result, list):
+            results['error_message'] = result[0]
+        else:
+            results['error_info'] = result['error']
```
sffc, line 199: Observation: sometimes you use `error_info` and other times you use `err_info`.

sffc (approved):

schema/schema_validator.py
sven-oly: I'll examine this.

PR #164, opened 2024-01-19
Update versions: CPP to 74.2, Node to 21.6.0 (ICU 74.1)

sven-oly:


sven-oly:
This updates the CPP testing version to 74.2.
It also uses the latest Node 21.6.0 with ICU 74.1

sven-oly:
PTAL to move this along. We should be using the newest version of Node.

To record the extended need, I just posted [PR# 168](https://github.com/unicode-org/conformance/issues/168).

sffc (approved):

PR #165, opened 2024-01-19
Parallelize verifier and separate verify_plan.py

sven-oly:
This involves moving much functionality into the VerifyPlan class in the file verify_play.py. It reorganizes the verifier object, simplifying and removing unneeded functions and data.

This also allow running multiple verify steps in parallel, using a python multiprocessing pool. This reduces the time for processing the end-to-end script.


verifier/verifier.py
```diff
@@ -152,10 +134,10 @@ def setup_verify_plans(self):
             executor_list = self.options.exec
             test_list = self.options.test_type
 
+        # Generate a pla for each executor, test_type, and icu_version
```
echeran, line 137:
```suggestion
        # Generate a plan for each executor, test_type, and icu_version
```

echeran (approved):
LGTM

PR #166, opened 2024-01-19
Add ICU4J executor functionality for number formatter

echeran:


sven-oly (approved):

sven-oly:
I tried this and I see that 11 out of 101 tests in the small run are executed. This is a good start!

sven-oly:
About 90% of the tests aren't showing up in the results. Perhaps these cases are not returning anything in StdOut?

echeran:
> About 90% of the tests aren't showing up in the results. Perhaps these cases are not returning anything in StdOut?

The latest commit in this PR addresses that. The reason for the problem is that the executor wasn't written robustly enough to handle when errors occur once the `test_type` is recognized. The fix is to ensure that a JSON still gets sent back in the error case that at least has the fields representing the error case.

sven-oly:
This is good progress!

PR #169, opened 2024-01-25
Fix errors in ICU4J number formatter

echeran:


sven-oly (commented):
This looks good so far!

echeran:
In local testing of 7334 test cases, 7307 pass, 17 fail, 10 errors.

sven-oly (approved):
Yay!

echeran:
@sffc When you can, take a look to see if the logic for applying the test options is correct according to the ICU4J NumberFormatter APIs. Also, see #171 for some of the issues encountered along the way.

PR #170, opened 2024-01-26
Characterize - add tristate controls

sven-oly:
This replaces the checkboxes with a tristate control for picking the set of tests to show. These are the options:
"?": does not select the characterization
checkmark: selects the characterization, intersecting with other selected options
"X": selects the characterization whose tests should be omitted from the set to display.

For example, for the following selections
- checkmark for characterization A
- checkmark for characterization C
- X for characterization D
the result is the the tests that are both in A and C, but not in D.

This PR also fixes the accidental omission of the timestamp for detailed reports.

### DIFF
--- a/verifier/detail_template.html
+++ b/verifier/detail_template.html
@@ -100,43 +100,48 @@ table th {
     // Stores info from the json data fetched asynchronously.
     let test_results = {
         'pass': {
-            json: null,
+            json: null,  // JSON characterized data
+            all_labels: new Set(),  // All the labels for this category
             count: 0,
             characterized: null,
-            check_boxes: [],
+            check_boxes: [],  // The widgets associated with characterized items
+            widget_label_sets: [],  // A parallel array of sets of labels for each.
             box_labels: [],
             selected_set: null
         },
         'fail': {
             json: null,
+            all_labels: new Set(),
             count: 0,
             characterized: null,
             check_boxes: [],
+            widget_label_sets: [], 
             box_labels: [],
             selected_set: null
         },
         'error': {
             json: null,
+            all_labels: new Set(),
             count: 0,
             characterized: null,
             check_boxes: [],
+            widget_label_sets: [],
             box_labels: [],
             selected_set: null
         },
         'unsupported': {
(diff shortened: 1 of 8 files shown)


verifier/detail_template.html
```diff
@@ -224,11 +224,41 @@
...
+      // Start the set of selected items.
+      test_data.selected_set = newSet;
+      } else {
+      // Update by intersection with this set.
+      test_data.selected_set =
+      new Set([...test_data.selected_set].filter((x) => newSet.has(x)));
+      }
+      }
+            else if (check_box.checked) {
+               // A checked box.
```
echeran, line 261: fix the indentation of this region

sven-oly: Fixed this. Also removed old code and revised a lot to be clearer. PTAL.

sven-oly:
PTAL. I think I've fixed the sets that are shown with selection of include / exclude.

sven-oly:
I cleaned up some parts of testreport.py to fix the characterizations and make somewhat better python code.

Ready for your review.

verifier/detail_template.html
```diff
@@ -150,145 +155,153 @@
...
+                // Update by removing all these labels, i.e., intersect with the inverse
+                for (let label of label_set) {
+                    excluded_set.add(label);
+                }
+            }
+        }
+        
+        // Remove the excluded values from selected set.
+        excluded_set.forEach((label) => {
+            test_data.selected_set.delete(label);
```
echeran, line 272: Comparing this line on 272 to line 260, it seems as if finding the intersection of the checked options is done one way (using `.filter((x) => label_set.has(x))`) that isn't followed when doing the intersection of the negation of the crossed/excluded options (using `.forEach((label) => {test_data.selected_set.delete(label);})`)

So long as it all works correctly, it's fine. And if you're using sets, and your code is always recomputing set intersections whenever the UI is updated after an updated selection of checks and crosses of key-value pairs, then I suppose it should all work.

echeran (approved):
LGTM

PR #177, opened 2024-02-14
Add collation testing for `dart_native`

mosuem:
Add testing for collation first to see how it goes, if it works then the other functions can be added. Same as Dart web is a wrapper around the browser's `Intl` object, Dart native is a wrapper around ICU4X.


sven-oly (approved):
LGTM. Thank you!

PR #178, opened 2024-02-16
Add optional test_types flag to testdata_gen script

mradbourne:
For certain use-cases, it is convenient to generate the JSONs for one test type rather than generating all tests each time (for example, when adding a new test type with a different schema). Passing a flag to the script is cleaner than making temporary code changes during development.

This PR adds that flag, which can be used as follows:

Run all tests (pre-existing behavior):
```
python testdata_gen.py --icu_versions icu74
```

Run a subset:
```
python testdata_gen.py --icu_versions icu74 --test_types number_fmt lang_names
```

### DIFF
--- a/testgen/testdata_gen.py
+++ b/testgen/testdata_gen.py
@@ -9,7 +9,7 @@ import multiprocessing as mp
 import os
 import re
 import requests
-from enum import StrEnum
+from enum import Enum
 
 reblankline = re.compile('^\s*$')
 
@@ -22,7 +22,7 @@ NUMBERS_TO_TEST = ['0', '91827.3645', '-0.22222']
 NUMBERFORMAT_LOCALE_INDICES = [3, 7, 11]
 
 
-class TestType(StrEnum):
+class TestType(str, Enum):
     NUMBER_FMT = 'number_fmt'
     COLLATION_SHORT = 'collation_short'
     LANG_NAMES = 'lang_names'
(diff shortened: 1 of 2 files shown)


echeran (approved):
LGTM!

sven-oly (approved):

sven-oly:
This is a good idea. We should also use this enum for the Python code in testdriver, schema, and verifier.

testgen/testdata_gen.py
```diff
@@ -9,7 +9,7 @@
 import os
 import re
 import requests
-import sys
+from enum import StrEnum
```
echeran, line 12: Merging is blocked because CI gives this error:
```
...
ImportError: cannot import name 'StrEnum' from 'enum' (/usr/lib/python3.10/enum.py)
```
According to [this answer](https://stackoverflow.com/questions/44666136/importerror-cannot-import-name-intenum), maybe we need to include a test for the `enum` module similar to what we do for the C JSON parsing library in our setup script.

Our setup script assumes a Debian-based system (including Ubuntu), which is what Github Actions runs on. Our local systems happen to be Debian-based, too, so we've written the script in a way that works locally, as well.

echeran: Ah, okay, @sven-oly 's hypothesis was correct: [`StrEnum` was added to Python in 3.11](https://tsak.dev/posts/python-enum/), while the [Github Actions runner VM only has Python 3.10](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#language-and-runtime).

Based on the first link describing `StrEnum`, it seems like the better choice when available, but perhaps if we revert back to plain `Enum`, all we lose is a consistent output when stringifying the enum? If so, I think we can live with that for our purposes. What do you think, @mradbourne ?

PR #183, opened 2024-02-21
Add date/time, list format, relative date/time data. Some executors added in NodeJS and ICU4X.

sven-oly:


sven-oly:
Please take a look. Note that CPP datetime instances all fail now, and CPP doesn't have list format.

executors/cpp/datetime_fmt.cpp
```diff
@@ -0,0 +1,257 @@
...
+  //   OTHER?
+
+  string dateStyle_str;
+  string timeStyle_str;
+  string calendar_str;
+  string timezone_str;
+
+  string era_str;
+  string year_str;
+  string month_str;
```
sffc, line 75: Question: I don't see where you read these variables. Intentional?

executors/cpp/datetime_fmt.cpp
```diff
@@ -0,0 +1,257 @@
...
+  json_object_object_add(return_json, "label", label_obj);
+
+  UDate testDateTime;
+  // The type of conversion requested
+
+  // Prefer milliseconds as input.
+  json_object *input_millis = json_object_object_get(json_in, "input_millis");
+  if (input_millis) {
+    testDateTime = json_object_get_double(input_millis);
+  } else {
```
sffc, line 212: This is a bit messy; you should either have the timestamp or the ISO string but having both just creates confusion. I prefer the ISO string because it is less ambiguous.

executors/cpp/datetime_fmt.cpp
```diff
@@ -0,0 +1,257 @@
...
+      const string input_date_string = json_object_get_string(input_string_obj);
+      cout << "# date from input_string: " << input_date_string << endl;
+
+      UnicodeString date_ustring(input_date_string.c_str());
+
+      UnicodeString parse_skeleton = "YYYY-MM-DD HH:mm:ss";
+
+      DateFormat* dparser = DateFormat::createInstanceForSkeleton(parse_skeleton,
+                                                                  displayLocale,
+                                                                  status);
```
sffc, line 224: You shouldn't use an ICU4C DateFormat to parse an ISO-8601 string. You should be able to parse it directly or using a helper library such as `absl`.

testgen/generators/list_fmt_gen.js
```diff
@@ -0,0 +1,143 @@
...
+// Set up Node version to generate data specific to ICU/CLDR version
+// e.g., `nvm install 21.6.0;nvm use 21.6.0` (ICU 74)
+
+const fs = require('node:fs');
+
+const debug = false;
+
+const locales = ['und',
+                 'en-US', 'zh-TW', 'pt', 'vi', 'el', 'mt-MT', 'ru', 'en-GB',
+                 'bn', 'ar','mr', 'zu'];
```
sffc, line 17: Suggestion: include Spanish, which uses inflection to pick between `o` and `u` and between `y` and `e`, and make sure there is a test that exercises this code path.

executors/cpp/datetime_fmt.cpp
sven-oly: Good idea! I'll add this.

executors/cpp/datetime_fmt.cpp
sven-oly: OK, I'm removing the milliseconds as an output field.

sven-oly: Reversal - parsing the date isn't yet working, so I'm keeping the milliseconds for now.

executors/cpp/datetime_fmt.cpp
sven-oly: They will be used when I figure out the API to handle individual fields in CPP version of DateTime format.

executors/cpp/datetime_fmt.cpp
sven-oly: Parsing date string isn't yet working.

testgen/generators/list_fmt_gen.js
sven-oly: Good idea!  Added!

executors/cpp/datetime_fmt.cpp
sffc: Let's chat about how to use `absl` here, which I think is the right thing to do.

executors/cpp/datetime_fmt.cpp
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  json_object *date_skeleton_obj = json_object_object_get(json_in, "datetime_skeleton");
+  if (date_skeleton_obj) {
+    // Data specifies a date time skeleton. Make a formatter based on this.
+    string skeleton_string = json_object_get_string(date_skeleton_obj);
+
+    cout << "# Skelecton = " << skeleton_string << endl;
+
+    UnicodeString u_skeleton = skeleton_string.c_str();
```
echeran, line 179: Apparently, this isn't the way to convert C String -> UnicodeString. You should use [this UnicodeString constructor](https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1UnicodeString.html#a613a04db62048c98e32112e5864716db), and only if the text is ASCII. If the `char *` text is not ASCII, then you need to do a codepage conversion first before converting to UnicodeString.

(I asked Shane b/c I didn't see any `char *` arg `operator=()` for UnicodeString. He learned the above info from Markus, so you can double check with him to make sure that's correct.)

echeran (commented):
initial comments. will continue

executors/cpp/datetime_fmt.cpp
sven-oly: Fixed this!

executors/cpp/datetime_fmt.cpp
sven-oly: Thanks. Constructor is now in place.

sven-oly:
OK, this is moving along. I've got a form of skeletons (not debugged) for use in ICU4C. However, it's not yet using Calendar or Timezone, so most cases still fail.

sven-oly:
PTAL. Note that Date Time formatting still has substantial test failure due to calendar not handled correctly.

New test data generator uses Node's formatToParts to get test data that is compatible with ICU4J/C with NBSP.

schema/datetime_fmt/test_schema.json
```diff
@@ -0,0 +1,139 @@
...
+           "type": "string"
+         },
+         "input_millis": {
+           "description": "numeric value for time in milliseconds since January 1, 1970, UTC (the epoch)",
+           "type": "string"
+         },
+         "input_string": {
+           "description": "String in ISO 8601 form, e.g. YYYY-MM-DD hh:mm:ss.sss",
+           "type": "string"
+         },
```
sffc, line 38: Are all three of these fields valid input?

schema/list_fmt/test_schema.json
```diff
@@ -0,0 +1,114 @@
...
+   },
+   "source_file": {
+     "description": "origin of conformance data",
+     "type": "string"
+   },
+   "source_version": {
+     "description": "version of conformance data",
+     "type": "string"
+   },
+   "test_ environmment": {
```
sffc, line 25: Extra space character?

schema/datetime_fmt/test_schema.json
sven-oly: Removed input and input_millis fields.

schema/list_fmt/test_schema.json
sven-oly: Good catch! Fixed!

sffc (approved):

PR #188, opened 2024-03-01
Fix collation short's escaping of collation test data

sven-oly:
This reduces collation test failures significantly.


sven-oly:
Pushing this small fix that resolves a lot of collation failures

PR #189, opened 2024-03-01
Refactor testdata_gen

mradbourne:
The `generateData` class is large. As we add more test types, keeping these in a single file becomes less manageable. This PR moves the contents of this large class into multiple modules.

## Proposed file structure
- `testdata_gen.py` - Entry point, arg parsing, parallelization. No test-type-specific logic.
- `generators/` - Contains the generators for all the different test types.
  - `base.py` - The abstract class for each test type generator to extend. It contains shared logic for reading and writing files.
  - `collation_short.py` (etc) - Implements a new `CollationShortGenerator.process_test_data()` method using the pre-existing logic from `data_generator.processCollationTestData()`. It also contains the utility functions related only to collation_short. This is invoked from `testdata_gen.py` as before.
  - `lang_names.py` - As above, it implements `process_test_data()` using the pre-existing logic.
  - `likely_subtags.py` - As above, it implements `process_test_data()` using the pre-existing logic.
  - `number_fmt.py` - As above, it implements `process_test_data()` using the pre-existing logic.

## Changes are mainly structure, not logic
Aside from the re-written `generate_versioned_data()` function in `testdata_gen.py`, this PR largely moves around existing code. I used 'black' formatter during the refactor, which has automatically applied consistent indentation/quotes/etc but, for ease of review, I have left most logic as-is*.

*I did remove `insert_nonignorable_coll_descr()` which has no references and appears to be obsolete.


echeran:
@mradbourne There was a merge conflict (according to git), and I think that's only because git got confused because the order of the entries in the `TestType` enum class changed. I resolved the merge conflict from upstream `main` and pushed the change to your branch.

sven-oly (approved):
Thanks for working on this! Good stuff!

echeran (approved):

mradbourne:
@echeran It looks like resolving the merge conflict re-introduced some unwanted old code. I've raised #197 to tidy this up. Would you mind reviewing? Thanks.

PR #190, opened 2024-03-04
Fix collation escaping, removing double backslash

sven-oly:


sven-oly:
This re-implements a change that was lost in refactoring of testgen.

echeran (approved):

PR #194, opened 2024-03-08
Pin Dart version to stable 3.2.6 for uninterrupted development

sven-oly:


sven-oly:
This is no longer needed.

PR #195, opened 2024-03-08
Add schema for verification files

sven-oly:
Next, add code to check verify files against these schema.


sffc (approved):
Seems good, although the four verify_schema.json files are the same? Are they intended to be different?

PR #197, opened 2024-03-12
Remove unused generateData class from testdata_gen

mradbourne:
This deleted code was re-added when [resolving a merge conflict](https://github.com/unicode-org/conformance/pull/189/commits/ce291948ed063ec172569b192fd0d6f58769bb07).
Since it's no longer used, we should remove it.


echeran (approved):
rslgtm. thanks!

PR #200, opened 2024-03-14
ICU4J: move all to ICU 74 from ICU73

sven-oly:
This updates the tested version of ICU4J to 74, allowing full testing of likely subtags with the correct data version.

This brings ICU4J results into very close alignment with results from ICU4C 74.


echeran (approved):
LGTM. Thanks!

PR #217, opened 2024-03-28
Switch to icu74

mosuem:


sven-oly:
#385 
 adds testing in versions 74, 75, and 76. This PR is now obsolete.

PR #220, opened 2024-04-03
Plural rules data (cardinal, ordinal) + Node executor

sven-oly:
Does not yet include pluralRange tests


executors/node/executor.js
```diff
@@ -149,6 +153,10 @@ function parseJsonForTestId(parsed) {
     return testTypes.TestRelativeDateTimeFmt;
   }
 
+  if (testId == "list_fmt") {
```
echeran, line 156:
```suggestion
  if (testId == "plural_rules") {
```

executors/node/plural_rules.js
```diff
@@ -0,0 +1,55 @@
+// Tests Intl Locale for pluralrules.
+
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale
```
echeran, line 3:
```suggestion
// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/PluralRules
```

executors/node/plural_rules.js
```diff
@@ -0,0 +1,55 @@
...
+    }
+    let plural_type;
+    if (json['type']) {
+      plural_type = json['plural_type'];
+      test_options['type'] = plural_type;
+    }
+
+    const supported_locales =
+          Intl.PluralRules.supportedLocalesOf(locale, test_options);
+    console.log("supported: ", supported_locales);
```
echeran, line 28: This could get noisy. The list of supported locales won't change from one invocation to the next, so are we logging the same info for each test case?

testgen/generators/plurals.py
```diff
@@ -0,0 +1,152 @@
...
+    def text_to_samples(self, text):
+        # TODO!!! Get text string to samples, applying rules =, !=, mod, .., ~
+        # n, i, v, f, t, w
+
+        # FIrst try: get the numbers following "@"
+        pos_at = text.find("@")
+        text_after_at = text
+        if pos_at >= 0:
+            text_after_at = text[pos_at:]
+        samples = re.findall(r'(\d+\.?\d*)',  text_after_at)
```
sffc, line 142: Ok... that should mostly work but we'll probably need to properly process these sample strings at some point.

sffc (approved):

executors/node/plural_rules.js
```diff
@@ -0,0 +1,55 @@
...
+    if (json['sample']) {
+      sample = Number(json['sample']);  // Convert to a number
+    }
+    let plural_type;
+    if (json['type']) {
+      plural_type = json['plural_type'];
+      test_options['type'] = plural_type;
+    }
+
+    const supported_locales =
```
echeran, line 26: Is there a way in JS to store this value in a global scope so that it doesn't need to be redefined each time?

executors/node/plural_rules.js
```diff
@@ -0,0 +1,55 @@
...
+          Intl.PluralRules.supportedLocalesOf(locale, test_options);
+    console.log("supported: ", supported_locales);
+    if (!supported_locales.includes(locale)) {
```
echeran, line 29: If there is a way for `supported_locales` to be stored in a global scope, then you could also [convert it into a Set](https://stackoverflow.com/questions/28965112/javascript-array-to-set). I would expect a Set to have a faster lookup than an array, which would be O(n).

testgen/generators/plurals.py
```diff
@@ -0,0 +1,152 @@
...
+        if tests_range:
+            self.json_test["tests"].extend(tests_range)
+        if verifications_range:
+            self.json_verify["verifications"].extend(verifications_range)
+
+        # And save the results
+        self.saveJsonFile("plural_rules_test.json", self.json_test, 2)
+        self.saveJsonFile("plural_rules_verify.json", self.json_verify)
+
+        return result
```
echeran, line 71: newline after to separate functions with whitespace vertically

testgen/generators/plurals.py
```diff
@@ -0,0 +1,152 @@
...
+        tests = []
+        verifications= []
+
+        label_num = 0
+
+        # TODO: Check if file exists
+        try:
+            tree = ET.parse(filename)
+        except:
+            logging.info('No plurals file found: %s', filename)
```
echeran, line 92: Seems like an error rather than info.

testgen/generators/plurals.py
```diff
@@ -0,0 +1,152 @@
...
+
+reblankline = re.compile("^\s*$")
+
+class PluralGenerator(DataGenerator):
+    def plurals_descriptor(self):
+        # Create test list of all these
+        self.test_id = 'plural_rules'
+        self.categories = ['zero', 'one', 'two', 'few', 'many', 'other']
+
+        source_url = ['https://github.com/unicode-org/cldr/blob/main/common/supplemental/plurals.xml',
```
echeran, line 19: Actually, I think you want to replace the word `main` with the release label (ex: `release-41`) in order for the CLDR data to match the version of ICU that this version platform is based on, right?

So if we're talking about generating ICU 71 data, then CLDR version is 41, so we want `'https://github.com/unicode-org/cldr/blob/release-41/common/supplemental/plurals.xml'`, etc.

If we don't change the word `main` in those URLs, then I would predict that all of our generated data files would be identical across ICU versions, and the contents of those data files would depend on what CLDR looked like at the time that we ran the data generation step most recently.

echeran (changes requested):
PTAL at my comments so that we can preemptively fix things, or else the clarify misunderstandings on my part.

(The approach of fetching plural rule samples via CLDR XML is the correct one. The APIs in ICU that return the samples as a fixed decimal are internal APIs.)

testgen/generators/plurals.py
sven-oly: This is a good catch. This source_url field isn't used to fetch the data, but is just a descriptor. But it should be accurate. I'll try to update this.

testgen/generators/plurals.py
sven-oly: Changed to a warning. This is not fatal to the data generation, but it may be a useful warning.

executors/node/plural_rules.js
sven-oly: Supported locales isn't a set of all hose supported by the API, but rather it's those that are compatible with the locale(s) entered. E.g., in the case of "iw", it returns "he", a more modern code for Hebrew. In the case of "mo", it gives "ro" (Romanian) instead of "Moldovan". Today's changes mark these locales as "unsupported".

PR #222, opened 2024-04-08
ICU4C PluralRules - first implementation

sven-oly:


executors/cpp/plural_rules.cpp
```diff
@@ -0,0 +1,171 @@
...
+    // Get the number
+    string sample_string = json_object_get_string(sample_obj);
+
+    if (sample_string.find('.') != std::string::npos) {
+      // Convert into an integer, decimal, or compact decimal
+      input_double_sample = std::stod(sample_string);
+      input_is_double = true;
+      cout << "#    DOUBLE STRING " << input_double_sample << endl;
+    } else
+      if (sample_string.find('c') != std::string::npos) {
```
echeran, line 80: The test for compact decimal should come before the test of a decimal point because `1.2c6` is a valid compact decimal number, but `3.1415926` isn't.

executors/cpp/plural_rules.cpp
```diff
@@ -0,0 +1,171 @@
...
+  int u_strings_size = 0;
+  if (sample_obj) {
+    // Get the number
+    string sample_string = json_object_get_string(sample_obj);
+
+    if (sample_string.find('.') != std::string::npos) {
+      // Convert into an integer, decimal, or compact decimal
+      input_double_sample = std::stod(sample_string);
+      input_is_double = true;
+      cout << "#    DOUBLE STRING " << input_double_sample << endl;
```
echeran, line 78: Remove low level stdout printing statements, or else use a logging library and convert to debug level log line.

update: based on the TODO statements, it seems like this PR is meant to have a followup PR, and the stdout printing is a best approximation of debugging for development purposes.  If that is the case, then instead of removing the stdout printing prematurely, add a few TODO comments to remove the stdout printing.

executors/cpp/plural_rules.cpp
```diff
@@ -0,0 +1,171 @@
...
+    } else
+      if (sample_string.find('c') != std::string::npos) {
+        input_is_compact = true;
+      } else {
+        input_int_sample = std::stoi(sample_string);
+        input_is_integer = true;
+        cout << "#    INT STRING " << input_int_sample << endl;
```
echeran, line 85: Ditto

executors/cpp/plural_rules.cpp
```diff
@@ -0,0 +1,171 @@
...
+  if (type_string == "cardinal") {
+    plural_type = UPLURAL_TYPE_CARDINAL;
+  } else
+    if (type_string == "ordinal") {
+      plural_type = UPLURAL_TYPE_ORDINAL;
+    } else {
+      // TODO: Report and error.
+    }
+
+  cout << "# Plural type: " << type_string << " " << plural_type << endl;
```
echeran, line 106: Ditto, here and throughout.

Thought: maybe there's a good logging library for C++ that we can "vendor in"?

executors/cpp/plural_rules.cpp
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  // TODO: distinguish between ints, doubles, and compact values
+  UnicodeString u_result;
+  if (input_is_double) {
+    cout << "Double select " << input_double_sample << endl;
+    u_result = prules->select(input_double_sample);
+  } else
+  if (input_is_integer) {
+    cout << "Double select " << input_double_sample << endl;
```
echeran, line 126:
```suggestion
    cout << "Integer select " << input_int_sample << endl;
```

executors/cpp/plural_rules.cpp
```diff
@@ -0,0 +1,171 @@
...
+  PluralRules* prules = icu::PluralRules::forLocale(
+      display_locale,
+      plural_type,
+      status);
+
+  if (U_FAILURE(status)) {
+    json_object_object_add(return_json,
+                           "error",
+                           json_object_new_string("construct plural rules"));
+  }
```
echeran, line 117: question for here and throughout: whenever there is an error encountered, do you want to do an early return after adding the error field to the return JSON? There are multiple instances of encountering an error, and there is an early return in one of those cases, but not the others.

executors/cpp/plural_rules.cpp
sven-oly: Good catch!

sven-oly: And a value can be both a double and a compact. This will need more work to see if ICU4C can handle compact input.

executors/cpp/plural_rules.cpp
sven-oly: Fixing

executors/cpp/plural_rules.cpp
sven-oly: Thanks. I think this is better now.

echeran (commented):
Looks good. Just need a minor tweak on the conditional logic. 

I also would like advice from @sffc on handling compact number input, but LMK if you want to merge without that and treat that as a followup improvement.

sven-oly:
I'm not sure what to do with Compact vs. Decimal. Here are some options:
1. Mark a compact value as "unsupported".
2.  Change "c" in a numeric string to "e", then treat as a decimal.
3. Something else?

@sffc, please advise.

sven-oly:
I'd appreciate more guidance on how to handle compact. Can we submit and then fix?

sffc:
Please mark the compact values as "unsupported". I think we established that they are currently supported in ICU4X only.

executors/cpp/plural_rules.cpp
sven-oly: Update code, fixing many lint complaints.

sven-oly:
I will set compact to unsupported.

sven-oly:
PTAL: linted...

sven-oly:
PTAL so we can add this component to ICU4C.

sffc (approved):

PR #227, opened 2024-05-14
Add icu75 for NodeJS and ICU4C execution

sven-oly:
Uses ICU75.1 and CLDR 45 data. Now there will be 7 versions of NodeJS in the data.


sven-oly:
PTAL. I'd love to get this checked in soon.

testdriver/datasets.py
```diff
@@ -208,7 +208,8 @@ class ParallelMode(Enum):
   ParallelByLang = 2
 
 class NodeVersion(Enum):
-  Node20 = "20.1.0"
+  Node22 = "22.1.0"
+  Node20 = "21.6.0"
```
echeran, line 212: Why does the `Node20` string begins with "21"? I don't see any usages of `Node22` in this PR, for example, so is that addition necessary, or can we remove `Node22`. And for that matter, where else does `NodeVersion` get used, or can we remove that altogether?

testgen/generators/common.js
```diff
@@ -0,0 +1,16 @@
+
+function xsample_tests(all_tests, run_limit) {
```
echeran, line 2: Is the spelling of `xsample_tests` intentional?

testgen/generators/datetime_gen.js
```diff
@@ -260,6 +285,19 @@ function generateAll(run_limit) {
           all_options['timeZone'] = timezone;
         }
         if (number_system) {
+          try {
+            const common_number_systems = intl_locale.numberingSystems;
+            const supported_calendars = intl_locale.calendars;
+            // TODO: Check if the number system is supported in this locale.
```
echeran, line 291: Is it easy enough to implement the code for the TODO task re: the number system?

echeran (approved):

testgen/generators/common.js
sven-oly: Fixing this. Thx.

testdriver/datasets.py
sven-oly: Removing NodeVersion. Good catch!

testgen/generators/datetime_gen.js
sven-oly: Updated this. PTAL.

testgen/generators/plural_rules_gen.js
```diff
@@ -0,0 +1,143 @@
...
+
+const styles = ['long', 'short', 'narrow'];
+
+const lists = [
+  [],
+  ['cat'],
+  ['dog', 'cat'],
+  ['fish', 'mouse', 'parakeet'],
+  ['3 weeks', '2 days', '3 hours', '25 minutes']
+];
```
sffc, line 29: Double-check this file; it looks like it generates list data, but:

1. It is named `plural_rules_gen.js`
2. It is in a PR that should be a clean ICU 75 update

sffc (approved):

testgen/generators/plural_rules_gen.js
sven-oly: Removed this generator. Thx for catching this.

PR #228, opened 2024-05-14
Basic Lint on .cpp files

sven-oly:
This doesn't include analysis of includes being used, nor names of variables.


echeran:
Is this using an open source tool that we can plug into our CI so that we do have to remember do this manually? Once we start linting & reformatting, it helps minimize code diffs to keep it up consistently, for which automation is the only real way to achieve that consistently (& without mental burden to boot).

It looks like Google open sourced a tool `cpplint` and then stopped maintaining the open source tool, but other people in open source have picked it up and kept maintaining it. It looks like you can install it via `pip`.
https://github.com/cpplint/cpplint

sffc:
`clang-tidy` is the industry standard.

sven-oly:
PTAL soon. I've removed the util and capi files and revised the text.

I'm using  cpplint.py now. Clang_tidy at Google only works in google3 directories. I'll look into installing a public version locally.

echeran (approved):
LGTM. I will file an issue about getting clang-tidy to install locally & in CI

PR #229, opened 2024-05-15
ICU4X: Adding list format testing

sven-oly:
Add another field list_type. The field "type" in list_fmt data is now deprecated.


sven-oly:
Almost all work. Locale und fails.

sffc (approved):
Nice! Looks about right

sffc:
Please run `cargo fmt` to satisfy the CI.

PR #232, opened 2024-05-21
Add plural rules to CPP executor

sven-oly:
This updates the main.cpp to recognize the plural_rules test type.

### DIFF
--- a/executors/cpp/main.cpp
+++ b/executors/cpp/main.cpp
@@ -45,14 +45,16 @@ extern const string test_langnames(json_object *json_in);
 extern const string test_likely_subtags(json_object *json_in);
 extern const string test_list_fmt(json_object *json_in);
 extern const string test_numfmt(json_object *json_in);
+extern const string TestPluralRules(json_object *json_in);
 
-std::string supported_tests[6] = {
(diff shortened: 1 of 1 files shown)


executors/cpp/main.cpp
```diff
@@ -83,44 +77,41 @@ int main(int argc, const char** argv)
       // TODO: get from the array of supported tests
       json_object *tests_supported = json_object_new_object();
       json_object *test_array = json_object_new_array();
-      for (int index = 0; index < 4; index ++) {
-        json_object_array_add(test_array,
-                              json_object_new_string(supported_tests[index].c_str()));
+      for (int index = 0; index < 7; index ++) {
```
echeran, line 80: replace magic number with built in functions that you can combine for the length of an array:
https://stackoverflow.com/a/4108340/2077918

based on the other answers on the page, it looks like there are nicer versions for if we were using C++17 or C++20

sven-oly: Hmm, I thought I had replaced this with a size call. Will revisit!

echeran (approved):
LGTM

PR #233, opened 2024-05-23
Quick cpp fix

sven-oly:


sven-oly (commented):
This is a small and non-controversial change to fix lang_names in CPP

PR #235, opened 2024-05-29
Clang tidy - initial application

sven-oly:
This fixed a lot of code, especially unused variables.


executors/cpp/langnames.cpp
```diff
@@ -60,8 +56,7 @@ const string TestLangNames (json_object *json_in) {
 
   char test_result_string[1000] = "";
 
-  int32_t chars_out = testLang.extract(
-      test_result_string, 1000, nullptr, status);
+  testLang.extract(test_result_string, 1000, nullptr, status);  // ignore return
```
sffc, line 59: Does `.extract` NUL-terminate the string? (here and elsewhere)

sffc (approved):

PR #239, opened 2024-06-18
Add ICU4X 1.5

robertbastian:


sffc:
Can this be done in a way that reduces duplication especially in main.rs?

robertbastian:
with a lot of work sure. is it worth it?

sven-oly:
Thanks for working on this, Rob!

Question to Shane & Elango: any reason not to approve this? A new main.rs may be necessary.

executors/rust/1.5/src/main.rs
```diff
@@ -0,0 +1,144 @@
...
+                run_language_name_test(&json_info)
+            } else if test_type == "likely_subtags" {
+                run_likelysubtags_test(&json_info)
+            } else if test_type == "list_fmt" {
+                run_list_fmt_test(&json_info)
+            } else if test_type == "plural_rules" {
+                run_plural_rules_test(&json_info)
+            } else {
+                Err(test_type.to_string())
+            };
```
sffc, line 125: This kind of code is highly likely to go stale. It would be _nice_ to keep it in sync across versions. I don't immediately see why it needs to be different between 1.4 and 1.5?

.github/workflows/run-rust.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        icu4x-version: [ '1.3', '1.4' ]
```
Beyondyou92, line 14: []() icu4x- version: [ '1.3' , '1.4' ]

Beyondyou92 (approved):

robertbastian:
Maybe @sffc can update this to his satisfaction.

sffc:
It's a bit unfortunate that this has bitrot since I left my review in June/July

robertbastian:
It's very unfortunate that this was not merged in June.

sffc:
Replaced by #360

PR #240, opened 2024-06-21
Rust datetime fmt and updates to ICU4C and NodeJS

sven-oly:
This gets some datetime formatting working in ICU4X. The default timezone for test data is now UTC.

Still missing: DT Skeletons in ICU4X


executors/rust/1.3/src/datetimefmt.rs
```diff
@@ -0,0 +1,158 @@
...
+        date.day,
+        time.hour,
+        time.minute,
+        time.second,
+    )
+    .expect("Failed to initialize ISO DateTime instance.");
+    let any_datetime = datetime_iso.to_any();
+
+    // Testing with a default timezone
+    let time_zone = CustomTimeZone::utc();
```
sven-oly, line 128: Elango says: you should use time zone. You currently only default to UTC.

executors/rust/1.3/src/datetimefmt.rs
```diff
@@ -0,0 +1,158 @@
...
+        Err(e) => {
+            return Ok(json!({
+                "label": label,
+                "error_detail": format!("{option_struct:?}"),
+                "error_type": format!("Failed to create DateTimeFormatter instance: {e:?}"),
+            }));
+        }
+    };
+
+    // TODO: Use a skeleton.
```
sven-oly, line 145: Elango says: Rewrite or remove comment? Is there a way to use skeleta in date time formatter in versions < 2.0?

executors/node/datetime_fmt.js
```diff
@@ -25,6 +25,12 @@ module.exports = {
     }
     let return_json = {'label': label};
 
+    let timezone;
+    try {
+       timezone = test_options['time_zone'];
+    } catch {
+      timezone = options['timeZone'] = 'UTC';
+    }
     // Get the date from input milliseconds.
```
sffc, line 34: Your variable `timezone` is unused

executors/rust/1.3/Cargo.toml
```diff
@@ -13,11 +13,16 @@ serde = "1.0.171"
 serde_json = "1.0.100"
 rustc_version_runtime = "0.1.*"
     
-icu = { version = "~1.3", features = ["serde", "icu_compactdecimal", "icu_displaynames", "compiled_data"] }
+icu = { version = "1.3.2", features = ["serde", "icu_compactdecimal",
```
sffc, line 16: Issue: You need `~` dependencies because otherwise cargo will install the latest version, not 1.3

testgen/generators/datetime_gen.js
```diff
@@ -172,6 +261,29 @@ const dt_fields = {
...
+  try {
+    this_date = new Date(date_string);
+  } catch (error) {
+    console.log('##### new Date: \"%s\"%s, %s from %s', error, date_string, zdt);
+    return null;
+  }
+
+  return this_date;
+}
+
```
sffc, line 286: Please use the Temporal cookbook example on how to perform this conversion. https://tc39.es/proposal-temporal/docs/cookbook.html#temporal-types--legacy-date

`new Date(zdt.epochMilliseconds)`

sven-oly:
Data generation is updated along with ICU4C and NodeJS.

Setting the CustomTimeZone is not yet working, however.

Rust version datetime_fmt.rs  is not compiling - I need some help with this:
"rror: derive(ULE) must be applied to a #[repr(C, packed)] or #[repr(transparent)] type
  --> /usr/local/google/home/ccornelius/.cargo/registry/src/index.crates.io-6f17d22bba15001f/icu_casemap-1.3.2/src/provider/exceptions_builder.rs:47:1
   |
..."

This comes from tinystr, I think.

Also,

sffc:
@Manishearth

I didn't realize that the zerovec patch would break users of ICU4X 1.3. Do we need to release a patch on the 1.3 branch?

testgen/generators/datetime_gen.js
sven-oly: Done

sven-oly:
Please take a look.

Manishearth:
> I didn't realize that the zerovec patch would break users of ICU4X 1.3. Do we need to release a patch on the 1.3 branch?

Maybe if the conformance tests need it.

sffc:
For now I fixed it by pinning the lockfile to the yanked versions, but ideally we won't have to do that, and of course it prevents us from upgrading to a newer Rust version.

executors/node/datetime_fmt.js
sffc: ^ `timezone` is still not being used

executors/cpp/datetime_fmt.cpp
```diff
@@ -180,9 +185,10 @@ const string TestDatetimeFmt(json_object *json_in) {
     return json_object_to_json_string(return_json);
   }
 
-  if (tz) {
-    df->setTimeZone(*tz);
-  }
+  // !!! IS OFFSET ALREADY CONSIDERED?
+  // if (tz) {
+  //   df->setTimeZone(*tz);
+  // }
```
sffc, line 191: I think you need to set the `tz` somewhere, or else you won't be formatting in a time zone.

sffc (approved):
OK, I think, but you will probably get a bunch of C++ and JS failures if you ignore the time zone like you seem to be doing.

PR #246, opened 2024-07-01
Adding package.json with Temporal polyfill

sffc:


sven-oly (approved):
Thank you! We need to get this working in CI soon

PR #251, opened 2024-07-11
Testgen add pkg lock

sven-oly:


executors/rust/1.3/src/relativedatetimefmt.rs
```diff
@@ -0,0 +1,104 @@
+ // https://docs.rs/icu/1.3.2/icu/datetime/struct.DateTimeFormatter.html
```
sffc, line 1: What is this file? It is JS but it is in the Rust directory

sven-oly: Removed it!

sffc (approved):

PR #254, opened 2024-07-19
message_fmt2: Update error types in schema

catamorphism:
Update the error list in testgen_schema.json to match the schema in the MFWG spec repo,
https://github.com/unicode-org/message-format-wg/blob/main/test/schemas/v0/tests.schema.json

Also update function tests to match the schema (just copied the tests from the current MFWG spec repo).


mradbourne (approved):

echeran (approved):
LGTM. Thanks!

PR #255, opened 2024-07-19
message_fmt2: Add more properties to schema

catamorphism:
Add properties to schema that are used in ICU tests: 'char', 'line', 'comment', 'srcs', 'ignoreJava', and 'ignoreCpp'

### DIFF
--- a/schema/message_fmt2/testgen_schema.json
+++ b/schema/message_fmt2/testgen_schema.json
@@ -56,7 +56,6 @@
       ]
     },
     {
-      "$comment": "Note that in a given file, either all tests must use 'src', or all tests must use 'srcs'.",
       "anyOf": [
         {
           "properties": {
@@ -69,28 +68,22 @@
         },
         {
           "properties": {
-            "defaultTestProperties": {
-              "required": [
-                "srcs"
-              ]
-            }
-          }
-        },
-        {
-          "properties": {
              "tests": {
-               "type": "array",
-               "items": {
-                "required": [
-                  "src"
-                ]
+               "type": "string",
+                 "items": {
+                  "required": [
+                    "src"
+                  ]
+                 }
+             }
+          }
         },
         {
           "properties": {
             "tests": {
               "type": "array",
               "items": {
-                "required": [ "srcs" ]
+                "required": [ "src" ]
               }
             }
           }
@@ -131,9 +124,6 @@
         "src": {
           "$ref": "#/$defs/src"
         },
-        "srcs": {
-          "$ref": "#/$defs/srcs"
-        },
         "params": {
           "$ref": "#/$defs/params"
         },
@@ -164,9 +154,6 @@
         },
         "src": {
           "$ref": "#/$defs/src"
-        },
-        "srcs": {
-          "$ref": "#/$defs/srcs"
         },
         "params": {
           "$ref": "#/$defs/params"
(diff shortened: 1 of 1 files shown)


schema/message_fmt2/testgen_schema.json
```diff
@@ -68,12 +69,28 @@
...
+        },
+        {
+          "properties": {
+             "tests": {
+               "type": "array",
+               "items": {
                 "required": [
                   "src"
                 ]
+        },
```
mradbourne, line 87: The nesting doesn't look valid here

schema/message_fmt2/testgen_schema.json
```diff
@@ -67,30 +66,24 @@
...
-        },
         {
           "properties": {
              "tests": {
-               "type": "array",
-               "items": {
-                "required": [
-                  "src"
-                ]
+               "type": "string",
```
mradbourne, line 72: This "type" should always be "array" because it is describing the "tests" property.
(Since "items" is already a child, I suppose "type" is technically not required and can be removed if it's misleading). 

I think we only need this:
```jsonc
// ...
    {
      "anyOf": [
        {
          "properties": {
            "defaultTestProperties": {
              "required": [
                "src"
              ]
            }
          }
        },
        {
          "properties": {
            "tests": {
              "type": "array",  // For clarity only
              "items": {
                "required": [
                  "src"
                ]
              }
            }
          }
        }
      ]
    },
// ...
```

and the "src" string/array definition is all handled under "$defs".

schema/message_fmt2/testgen_schema.json
```diff
@@ -209,13 +196,17 @@
...
+        {
+          "description": "The MF2 syntax source.",
+          "type": "string"
+        },
+        {
+          "description": "The MF2 syntax source, as an array of strings to be concatenated.",
+          "type": "array",
+          "items": { "type": "string" }
+        }
+      ]
```
mradbourne, line 209: Nice!

schema/message_fmt2/testgen_schema.json
catamorphism: Fixed -- I also noticed that the second "properties" clause in the "anyOf" was duplicated for some reason (probably my mistake), so I removed that.

schema/message_fmt2/testgen_schema.json
```diff
@@ -68,14 +68,14 @@
         },
         {
           "properties": {
-            "tests": {
-              "type": "array",
-              "items": {
-                "required": [
-                  "src"
+             "tests": {
+               "type": "array", // For clarity only
```
mradbourne, line 72:
```suggestion
               "type": "array",
```
Ah, I added this in my other GitHub comment to help explain but we should avoid it in the schema since it's invalid JSON. Sorry for the confusion!

schema/message_fmt2/testgen_schema.json
echeran: Yep, this is good change, thanks. @mihnita was asking for something like this when he went ahead and wrote his [ICU4J unit tests as such](https://github.com/unicode-org/icu/blob/main/icu4j/main/core/src/test/resources/com/ibm/icu/dev/test/message2/icu-test-functions.json) (modulo a different field name), given that JSON doesn't give any support to make multi-line strings developer-friendly

echeran (approved):

schema/message_fmt2/testgen_schema.json
echeran: +1. The end-to-end CI check is failing unsurprisingly as a result during the initial step to validate the schemas. (FYI @sven-oly schema validation is doing it's job...)

catamorphism: Fixed.

PR #258, opened 2024-08-01
Adding ICU4J list format tests

sven-oly:
Thanks, Elango, for your help with this!


echeran (approved):
LGTM. If the CI job for end-to-end running works, that then implies that the unit tests pass.

PR #278, opened 2024-08-16
Adding ICU4C message format2 testing

sven-oly:
This executes MF2 tests in ICU75.1 and later.

There are some problems with the tests, but this is a start.

### DIFF
--- a/executors/cpp/message_fmt2.cpp
+++ b/executors/cpp/message_fmt2.cpp
@@ -73,6 +73,10 @@ const string TestMessageFormat2(json_object *json_in) {
     test_description_string = json_object_get_string(test_description_obj);
   }
 
+  // Start filling the return data.
+  json_object *return_json = json_object_new_object();
+  json_object_object_add(return_json, "label", label_obj);
+
   json_object* param_list_obj = json_object_object_get(json_in, "params");
 
   string params_name;
@@ -107,13 +111,16 @@ const string TestMessageFormat2(json_object *json_in) {
         u_params_value = params_value.c_str();
 
         argsBuilder[u_params_name] = Formattable(u_params_value);
+      } else {
+        // Bail out due to incomplete parameter
+        if (check_icu_error(
+                U_MESSAGE_PARSE_ERROR,
+                return_json, "incomplete param name / value")) {
+          return json_object_to_json_string(return_json);
+        }
       }
-    }
-  }
-
-  // Start filling the return data.
-  json_object *return_json = json_object_new_object();
-  json_object_object_add(return_json, "label", label_obj);
+    }  // End of parameter loop
+  }  // End parameter list processing
 
   UParseError parseError;
   UErrorCode errorCode = U_ZERO_ERROR;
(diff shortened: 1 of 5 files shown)


executors/cpp/message_fmt2.cpp
```diff
@@ -0,0 +1,180 @@
+/********************************************************************
+ * testing message format 2 in C++
+ */
+
+#include <unicode/utypes.h>
+
+#include <json-c/json.h>
+
+// This API was added in ICU75.1
+#if U_ICU_VERSION_MAJOR_NUM >= 75
```
echeran, line 10: nit: add a newline after

executors/cpp/message_fmt2.cpp
```diff
@@ -0,0 +1,180 @@
...
+  } else {
+    // It worked!
+    json_object_object_add(return_json,
+                           "result",
+                           json_object_new_string(test_result));
+  }
+
+  return json_object_to_json_string(return_json);
+}
+#endif
```
echeran, line 180: nit: add a newline (so that the `#endif` doesn't blend into the background) and a comment indicating the condition of the if statement, since it encapsulates so much code
```suggestion
}

#endif  // U_ICU_VERSION_MAJOR_NUM >= 75
```

executors/cpp/message_fmt2.cpp
```diff
@@ -0,0 +1,180 @@
...
+      json_object* param_obj = json_object_array_get_idx(param_list_obj, i);
+
+      json_object *params_name_obj = json_object_object_get(param_obj, "name");
+      if (params_name_obj) {
+        params_name = json_object_get_string(params_name_obj);
+        u_params_name = params_name.c_str();
+      }
+      json_object *params_value_obj =
+          json_object_object_get(param_obj, "value");
+      if (params_value_obj) {
```
echeran, line 97: suggestion: if you move `json_object *params_value_obj = ...` up to where `json_object *params_name_obj = ...` is, then you can combine `if (params_name_obj) {` and `if (params_value_obj) {`, and then add an `else` branch saying `continue`.

Right now, if either the name or the value is empty, then it seems meaningless and potentially confusing/incorrect to try to add the name+value pair to the args builder.

echeran (commented):
minor code changes requested, otherwise looks good

executors/cpp/message_fmt2.cpp
```diff
@@ -0,0 +1,180 @@
...
+        return_json,
+        "error_detail", json_object_new_string(error_name));
+    return json_object_to_json_string(return_json);
+  }
+
+  int32_t chars_out;  // Extracted characters from Unicode string
+  char test_result[1000] = "";
+
+  // Get the resulting value as a string
+  result_ustring.extract(test_result, 1000, nullptr, errorCode);
```
echeran, line 168: The API for UnicodeString -> UTF-8 std::string is `UnicodeString::toUTF8String(std::string)`.

echeran: https://unicode-org.github.io/icu-docs/apidoc/released/icu4c/classicu_1_1UnicodeString.html#aa2a91788b598b9aeb88ab4123a047c03

executors/cpp/message_fmt2.cpp
sven-oly: OK

executors/cpp/message_fmt2.cpp
sven-oly: OK

executors/cpp/message_fmt2.cpp
sven-oly: OK

echeran (approved):
LGTM!

PR #282, opened 2024-08-19
Clean up .cpp files, revising error reporting, adding util.h

sven-oly:


executors/cpp/datetime_fmt.cpp
```diff
@@ -205,7 +202,7 @@ const string TestDatetimeFmt(json_object *json_in) {
 
     // SimpleDateFormat can't parse options or timezone offset
     // First, remove options starting with "["
-    std:size_t pos = input_date_string.find("[");
+ std:size_t pos = input_date_string.find("[");
```
echeran, line 205: indentation seems off

executors/cpp/coll.cpp
```diff
@@ -204,89 +179,91 @@ const string TestCollator(json_object *json_in) {
...
-      us1.extract(char_out1, 1000, nullptr, status);  // ignore result
-      if (U_FAILURE(status)) {
-        test_result = error_message;
-        json_object_object_add(
-            return_json,
-            "error", json_object_new_string("error extracting us1"));
+    // Check unescaped versions.
+    char char_out1[1000] = "";
+    char char_out2[1000] = "";
+    us1.extract(char_out1, 1000, nullptr, status);  // ignore result
```
echeran, line 224: Any potential for `us1.extract(...)` and `us2.extract(...)` to use the `UnicodeString::toUTF8String(std::string)` API instead?

executors/cpp/coll.cpp
```diff
@@ -29,6 +29,8 @@
 #include <iostream>
 #include <string>
 
+#include "./util.h"
```
echeran, line 32: question for here & through the other files: can we change  `"util.h"` instead of `"./util.h"`? If so, it'll appear more clearly that way.

echeran (commented):
a few suggestions, mostly looking good

executors/cpp/datetime_fmt.cpp
sven-oly: This should be std with two colons!

executors/cpp/coll.cpp
sven-oly: The simple linter recommened using "./"

executors/cpp/coll.cpp
sven-oly: And in other files, too!

echeran (approved):
LGTM

Issue #250, opened 2024-07-11
Remove or update "Node.js 18 sample tests" - deprecated version of Node.


sven-oly:
[Run Node.js 18 sample tests](https://github.com/unicode-org/conformance/actions/runs/9899170766/job/27347576842)

The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

Run Node.js 18 sample tests
The following actions uses Node.js version which is deprecated and will be forced to run on node20: actions/checkout@v2, actions/setup-node@v3. For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

sffc:
@sven-oly suggests deleting this test.

sven-oly:
Removed in PR #284

PR #284, opened 2024-08-19
Remove "Node.js 18 sample tests", issue #250

sven-oly:
This is obsolete and doesn't need to run anymore. It also was using an older version of Node that is no longer supported in the workflow.


sven-oly:
Please take a look at this PR. It's waiting for the check "Run Node.js 18 sample tests" to complete, but this PR removes that workflow item.

echeran (approved):
LGTM

PR #285, opened 2024-08-20
Rust rdtf numeric - supports auto and always options

sven-oly:
This handles "yesterday" as well as "1 day ago".


echeran (approved):
LGTM

PR #289, opened 2024-08-21
ICU4X rel date time -  non-Latn number system --> unsupported

### LINKED ISSUE #287: ICU4X RelativeDateTimeFormatter: mark numeringSystem as unsupported
This code should be added to the relative date time formatter file:

 // Update when ICU4X supports non-Latn numbering systems
    // if option_struct.numbering_system.is_some() {
    //     let numbering_system = option_struct.numbering_system.as_ref().unwrap();

    //     if numbering_system != "latn" {
    //         return Ok(json!({
    //             "error": "Number system not supported",
    //             "error_msg": numbering_system,
    //             "label": label,
    //             "unsupported": "non-Latn numbering system not implemented",
    //         }));
    //     }
    // }

sven-oly:
This should be removed when numbering systems are better supported.

close #287

### DIFF
--- a/executors/rust/1.3/src/relativedatetime_fmt.rs
+++ b/executors/rust/1.3/src/relativedatetime_fmt.rs
@@ -132,9 +132,21 @@ pub fn run_relativedatetimeformat_test(json_obj: &Value) -> Result<Value, String
 
     let locale_json_str: &str = json_obj["locale"].as_str().unwrap();
     let mut locale_str: String = locale_json_str.to_string();
+
     if numbering_system_str.is_some() {
-        locale_str =
-            locale_json_str.to_string() + "-u-nu-" + &numbering_system_str.as_ref().unwrap();
+       let numbering_system = numbering_system_str.as_ref().unwrap();
+       locale_str =
+            locale_json_str.to_string() + "-u-nu-" + numbering_system;
+
+        // TODO: update when ICU4X supports numbering systems.
+        if numbering_system != "latn" {
+            return Ok(json!({
+                "error": "Number system not supported",
+                "error_msg": numbering_system,
+                "label": label,
+                "unsupported": "non-Latn numbering system",
+            }));
+        }
     }
 
     let lang_id = if let Ok(lc) = locale_str.parse::<Locale>() {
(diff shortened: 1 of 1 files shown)


executors/rust/1.3/src/relativedatetime_fmt.rs
```diff
@@ -132,9 +132,21 @@ pub fn run_relativedatetimeformat_test(json_obj: &Value) -> Result<Value, String
 
     let locale_json_str: &str = json_obj["locale"].as_str().unwrap();
     let mut locale_str: String = locale_json_str.to_string();
+
     if numbering_system_str.is_some() {
-        locale_str =
-            locale_json_str.to_string() + "-u-nu-" + &numbering_system_str.as_ref().unwrap();
+       let numbering_system = numbering_system_str.as_ref().unwrap();
+       locale_str =
+            locale_json_str.to_string() + "-u-nu-" + numbering_system;
```
echeran, line 139: The proper way to modify a locale is not to do string concatenation b/c what if the locale coming in already has that specified?

We should probably use the `icu::locid` crate for inspecting and setting Unicode [programmatically via the API](https://docs.rs/icu/1.3.0/icu/locid/extensions/unicode/struct.Keywords.html).

executors/rust/1.3/src/relativedatetime_fmt.rs
```diff
@@ -132,9 +132,21 @@ pub fn run_relativedatetimeformat_test(json_obj: &Value) -> Result<Value, String
 
     let locale_json_str: &str = json_obj["locale"].as_str().unwrap();
     let mut locale_str: String = locale_json_str.to_string();
+
     if numbering_system_str.is_some() {
-        locale_str =
-            locale_json_str.to_string() + "-u-nu-" + &numbering_system_str.as_ref().unwrap();
+       let numbering_system = numbering_system_str.as_ref().unwrap();
+       locale_str =
```
echeran, line 138: Also, FYI, `cargo fmt` needs to be applied b/c the formatting check is not happy about something somewhere...

sven-oly: OK

echeran (approved):
LGTM

PR #290, opened 2024-08-22
Lang name generator from cldr and using it

sven-oly:
This now uses the new locale name data in ICU73 and newer instead of the old data.

Executors now can use "dialect" option in ICU4C, ICU4J, ICU4X, and Node JS. However, there are many test failures and errors reported now. Next, evaluate why these differ.


executors/cpp/localedisplaynames.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+  // The locales's name to be displayed.
+  json_object *language_label_obj = json_object_object_get(
+      json_in, "language_label");
+  string language_label_string = json_object_get_string(language_label_obj);
+
+  // Either standard or dialect names for the locale.
+  json_object *language_display_obj = json_object_object_get(
+      json_in, "languageDisplay");
+  string language_display_string = json_object_get_string(language_display_obj);
```
echeran, line 41: rename these local variables from `language_*` to `locale_*`? you did above, just didn't continue that renaming here.

executors/icu4j/74/executor-icu4j/src/main/java/org/unicode/conformance/testtype/langnames/LangNamesDisplayOptions.java
```diff
@@ -0,0 +1,16 @@
+package org.unicode.conformance.testtype.langnames;
+
+public enum LangNamesDisplayOptions {
+  STANDARD,
+  DIALECT;  // DIALECT_NAMES is the ICU4J enum
```
echeran, line 5:
```suggestion
  DIALECT;  // ULDN_DIALECT_NAMES is the ICU4C enum
```

executors/icu4j/74/executor-icu4j/src/main/java/org/unicode/conformance/testtype/langnames/LangNamesTester.java
```diff
@@ -24,6 +24,12 @@ public ITestTypeInputJson inputMapToJson(Map<String, Object> inputMapData) {
     result.language_label = (String) inputMapData.get("language_label", null);
     result.locale_label = (String) inputMapData.get("locale_label", null);
 
+    String lang_display_string = (String) inputMapData.get("languageDisplay", null);
+    if (lang_display_string == null || lang_display_string.equals("standard")) {
```
echeran, line 28: Instead of keeping this is as a `String` and doing a string comparison, you should use the string->enum parsing function that you created in the enum class. That function also defines and handles the default.

executors/icu4j/74/executor-icu4j/src/main/java/org/unicode/conformance/testtype/langnames/LangNamesDisplayOptions.java
sven-oly: Done

executors/icu4j/74/executor-icu4j/src/main/java/org/unicode/conformance/testtype/langnames/LangNamesTester.java
sven-oly: Done

executors/cpp/localedisplaynames.cpp
sven-oly: Done

echeran (approved):
LGTM

Issue #257, opened 2024-07-25
Add containerization


mradbourne:
### Challenges
- Setup scripts target Ubuntu, which means a setup overhead for local development on Mac and Windows.
- Conformance testing requires the configuration of multiple platforms. Lack of isolation around this causes overhead for local development, especially if the developer is using the same tooling for multiple projects.
- The dependency specification is not centralized. For example, [dependency installation commands are included in the test generation scripts](https://github.com/unicode-org/conformance/blob/9e2e8eab8518065647df5a275e9209e7da81093f/testgen/generators/datetime_fmt.py). This makes getting a clear overview of dependencies more difficult.

### Proposed solution
Use Docker to containerize the environment and specify all dependency installation and setup in the Docker image definition.
Provide a Python CLI as the main entry point for the project, which abstracts the Docker complexity.

Containerization can be introduced incrementally:
1. Provide a Dockerfile that matches the environment specified in `.github/workflows/e2e.yaml` and `setup.sh`.
2. Create a simple Python script to easily run existing test scripts in a Docker container. For example:
```sh
$ python conformance.py run genData100.sh
```
3. Move all existing installation commands into the Dockerfile.
4. Allow run configuration files to be passed to the new Python CLI (i.e. `run_config.json` support).
5. Use the same Docker image and Python CLI in GitHub CI.
6. Move all existing test scripts into the Python CLI and add params or flags for executors and ICU versions. For example:
```sh
$ python conformance.py test --icu_versions=icu74 --test_types=rust
```

### Impact
This solution would introduce Docker as a dependency for the project.
Consolidating the dependency list and setup into the Dockerfile would simplify the project but it would also make Docker an integral part of the setup. There is room in this proposal to keep an uncontainerized version of the setup at the expense of some added complexity.

mradbourne:
If this seems like a good approach, I would be happy to work on implementing the solution.
All thoughts and feedback welcome.
cc @echeran

echeran:
That sounds great to me! I like the listing of the separate incremental steps to make this change (easier to review & adapt to). Overall, this ought to make the project easier to use and maintain.
FYI @sffc @sven-oly

sffc:
Things that are very architecture-specific, such as the C++ executor, would be good to containerize. _Maybe_ some of the other executors like Rust or Dart, especially if they depend on very specific compiler setups.

[...]

PR #293, opened 2024-08-23
Updating logging output - simplifying output

sven-oly:


echeran (approved):
LGTM!

I wish that the `logrotate` commands that I set up would work so that it's easier to sift through the debugging log files rather than having them accrete into one large file. Maybe that's just a nice to have thing, and in the meantime, opening the log files in VS Code allows us a colorized view with a real-time update like `tail -f` provide. If #257 happens, that might the first step to being able to figure out log rotation predictably. But I digress...  Accurately setting log levels is always desirable.

PR #301, opened 2024-09-17
Remove 'und' as a locale in testing

sven-oly:
This reduces test failures because the "und" locale can produce platoform-specific results.


echeran (approved):
LGTM

PR #305, opened 2024-09-18
Fsync for file output

sven-oly:
This make sure that the output is synchronized with parallel execution.


sven-oly:
This is an easy one!

schema/check_generated_data.py
```diff
@@ -91,6 +91,7 @@ def main(args):
         output_filename = os.path.join(test_data_path, 'test_data_validation_summary.json')
         file_out = open(output_filename, mode='w', encoding='utf-8')
         file_out.write(summary_data)
+        os.fsync(file_out)
```
echeran, line 94: Here and below: follow [instructions](https://docs.python.org/3/library/os.html#os.fsync) for `os.fsync()` to include the appropriate argument and additionally do `os.flush()` prior to the `fsync`, since you are getting a file object from `os.open()`.

sven-oly:
Abandoning this PR because it's not needed.

PR #315, opened 2024-09-26
Set ICU4J executor version number based on ICU version

### LINKED ISSUE #271: Fix ICU version number reported by ICU4J executor
Previously, the ICU4J executor only supported one ICU version, so the ICU version was hardcoded. Now that the ICU4J execcutor can support multiple ICU versions, it should report the ICU version correctly.

echeran:
Fixes #271


sven-oly (approved):
Thank you!

PR #319, opened 2024-10-02
Node errors for locale display names to 'unsupported'

### LINKED ISSUE #318: Locale display names in Node: change problems with RangeError to "unsupported"
These errors are actually failure to support locale options.

In addition, "root" isn't supported.

sven-oly:
This fixes all the "error" cases in NodeJs for locale display names, moving them to "unsupported". This works for unsupported locale options and also the "root" locale.

Fixes #318


sven-oly:
Fixes #318

executors/node/localedisplaynames.js
```diff
@@ -35,6 +35,12 @@ module.exports = {
         "error_type": "unsupported",
         "error_retry": false  // Do not repeat
       };
+      if (error instanceof RangeError) {
+        // The locale can't be handled for some reason!
+        outputLine["error_type"] = 'unsupported';
+        outputLine["unsupported"] = error.toString();
+        outputLine["error_detail"] = 'unsupported locale';
+      }
```
sffc, line 43: `error instanceof RangeError` seems a bit broad. I suggest a narrower case for returning unsupported.

sven-oly: Hmm, there are two types of failures right now. Here's what I see being caught as exceptions:

1. Unexpected options in the locale spec, e.g., "label":"0009","language_label":"**en-u-nu-deva-t-de**","locale_label":"en"

  --> "RangeError: invalid_argument"

2. "root" locale --> "RangeError: invalid_argument" in the constructor.

Should either of these be "unsupported"?

sven-oly: I think classifying these as "unsupported" is correct.  This will resolve a large number of "errors" --> "unsupported", i.e.,

Report created: 2024-10-07 10:36

1,990 attempted. Pass: 1,460, Fail: 214, **Errors: 316,** Unsupported: 0

sven-oly: PTAL!

sven-oly:
I've updated this to categorize unsupported locale options as "Unsupported". This reduces errors and failures.
image
image

I think this is worth submitting. PTAL.

sven-oly:
Please review again!

sffc (changes requested):
This is not ready. The problem is that the locale is "root" which is not a valid locale.

sven-oly:
Root locale was removed in PR #341

sven-oly:
This now checks the requested locale for validity as a locale. It also recognizes problems with getting the result as "unsupported" instead of test failures or errors.

sven-oly:
Note that there are still test failures in all platforms for a variety of reasons.

sven-oly:
PTAL. It would be good to resolve this issue #318 .

sven-oly:
I would like to merge this. It solves a problem that is in Q4 P1 milestones.

sven-oly:
PTAL

sffc (approved):

PR #320, opened 2024-10-02
Adding ICU67 for some ICU4C and NodeJS components

sven-oly:
This adds the ability to test an older version of ICU with ICU4C and NodeJS.


sven-oly:
Support for these older ICU versions is not required.

PR #321, opened 2024-10-03
Nodejs collation tests: Change "fail" to "unsupported" when a rule or complex comparison is used.

sven-oly:
These tests should be flagged as unsupported because the ECMAScript implementation does not support custom rules or other forms of comparison, i.e., levels of comparison.


sven-oly:
Note that the same should be done in Dart Web.

And rules should be implemented in ICU4 C, J, and X, too.

run_config.json
```diff
@@ -1,113 +1,4 @@
-[
```
sffc, line 1: Issue: `run_config.json` should be changing in this PR

executors/node/collator.js
```diff
@@ -39,13 +44,24 @@ module.exports = {
         result_bool = false;
       }
       outputLine = {'label':json['label'],
-                    'result': result_bool,
-                    'compare_result': compared,
                    }
-
-      if (result != true) {
+      if (result == true) {
```
sffc, line 48: Issue: if `result == false`, it should be returned as a potential test failure. Only if the options are unsupported should an unsupported status be returned.

run_config.json
sven-oly: I'll remove this file.

executors/node/collator.js
sven-oly: OK. I see that there's an ECMAScript option for sensitivity. I'll see if that fits with any of the "compare_type" options.

And should we set all tests that have explicit 'rules' to be 'unsupported'?

sffc: You mean like this test?

`{"test_type": "collation_short", "label":"00003","s1":"9","s2":"\u0000","compare_type":"&lt;1","test_description":" simple CEs &amp; expansions","rules":"&amp;\\x01&lt;&lt;&lt;\\u0300&amp;9&lt;\\x00&amp;\\uA00A\\uA00B=\\uA002&amp;\\uA00A\\uA00B\\u00050005=\\uA003","hexhash":"d439f742bdde4b51012ee05519dfde424e25334a"}`

Yeah, explicit "rules" are unsupported.

echeran, line 48: Question about initializations of `result` and `result_bool`, even though they're not in this PR: shouldn't `result` and `result_bool` always be identical, by definition?

If so, can we just use one of them instead of both, when it comes to the code here in the body of this if statement?

echeran, line 48: Leaving this comment here since I can't on lines above: aren't lines 43-45 redundant? We should be able to remove them b/c the logic is already covered by line 41, right?

sven-oly: Yes, I'll remove result_bool.

sffc (approved):

PR #328, opened 2024-10-10
Update handling of subdirectory creation

### LINKED ISSUE #327: Fix handling of existing directories to not fail
testdriver/testplan.py and verifier/verifier.py now fail if creating a subdirectory doesn't work. This should be changed to be more robust.

sven-oly:
Fixes #327


echeran (approved):
LGTM

PR #329, opened 2024-10-11
Datetime date from CLDR data

sven-oly:
Ready for review.


sven-oly:
I need some help converting these date/time data into UTC "Z" format. Then we can compute the offset seconds too!

sven-oly:
Note that the CLDR expected test results may be incorrect in some cases:
1. date/time styles full/full have "at" between the date and time output in ICU4C. The expected result isn't correct.

Expected: July 2, 2001, 5:44:15 PM GMT+4:30
Actual:      July 2, 2001 at 5:44:15 PM GMT+4:30

  {
    "dateLength": "long",
    "calendar": "gregorian",
    "locale": "en-US",
    "input": "2001-07-02T17:44:15+04:30[Asia/Tehran]",
    "expected": "July 2, 2001, 5:44:15 PM GMT+4:30"
  },

2. In some cases , there's GMT in the expected
March 7, 2024, 11:30:01 AM GMT+3:30
March 7, 2024 at 11:30 AM

  {
    "dateLength": "long",
    "calendar": "gregorian",
    "locale": "en-US",
    "input": "2024-03-07T11:30:01+03:30[Asia/Tehran]",
    "expected": "March 7, 2024, 11:30:01 AM GMT+3:30"
  },

  {
    "timeLength": "long",
    "calendar": "gregorian",
    "locale": "en-US",
    "input": "2024-03-07T11:30:01+03:30[Asia/Tehran]",
    "expected": "March 7, 2024, 11:30:01 AM GMT+3:30"
  },

sven-oly:
This now uses simple logic to identify the "known issue" of " at" replacing "," in some cases. Similarly, some other substitutions are caught.

Note also that the test data shows an ASCII comma as expected instead of the Arabic comma. These show up as test failures, but maybe it's a bug in the CLDR data.

verifier/check_known_issues.py
```diff
@@ -117,6 +121,32 @@ def unsupported_unit_quarter(test):
...
 
 
+def dt_check_for_alternate_long_form(actual, expected):
+    # For datetime_fmt, check if the word for "at" is inserted
+    # in the actual vs. the expected.
+    at_inserts = ['at ', 'এ ', 'lúc ', ',', ' এ']
+    replacements = {',': ' at', '،': ' في',
+                    }
+
+    sm = SequenceMatcher(None, expected, actual)
```
sffc, line 131: This is a much more complicated known-issue catcher than I was expecting. I was expecting that we would put a blanket "known issue" on all failing inputs where the dateTimeFormatType is "standard".

sven-oly: OK, I'll take a look at the failures to see if this will capture the substitution of "," with " at" and similar replacements in other languages.

sven-oly:
Categorize failing date time tests with 'dateTimeFormatType" == 'standard' to be a  "Known Issue".

sven-oly:
Ready for another look!

sven-oly:
This classifies most of the date time format problems as "known issue" except for the different beween ASCII comma and arabic comma.

verifier/testreport.py
```diff
@@ -731,7 +731,8 @@ def characterize_datetime_tests(self, test_list, results):
             if 'skeleton' in  test['input_data']:
                 skeleton_str = 'skeleton: ' + test['input_data']['skeleton']
                 results.setdefault(skeleton_str, []).append(label)
-
+            if 'dateTimeFormatType' in test:
+                results.setdefault('dateTimeFormatType: ' + test['dateTimeFormatType'], [])
```
echeran, line 735: do you need to chain a `.append(label)` to this operation? otherwise, as it is, `.setdefault()` doesn't actually do anything to the map, it just gives a default value operate on when the key doesn't yet exist

echeran (approved):
LGTM

PR #330, opened 2024-10-15
Unify CLI - part 1

mradbourne:
## Summary
This PR introduces a new `./conformance` script, which acts as a single CLI entry point to the different existing features of the codebase. This gives some advantages:

- It enables easier triggering of compound commands in future. (e.g. "generate, run and verify all ICU73 tests")
- It allows easier sharing of code such as CLI option definitions.
- It will afford easier integration of containerization or remote running in future. This could be used to test more platforms from a single host OS. It could also reduce setup complexity.
- It could offer improved self-documentation of all Conformance features in a single place (e.g. `./conformance --help`).

## Existing functionality is maintained
To minimize disruption, it maintains existing script entry points and keeps the CLI options the same. e.g.:
```sh
# The new command
./conformance test --icu_version=icu74 --test_type=message_fmt2 --exec=node        
# works the same as the old command
(cd testdriver && python3 testdriver.py --icu_version=icu74 --test_type=message_fmt2 --exec=node)
```

The intention is that Conformance users can begin using the unified CLI but, when it proves to be completely fit-for-purpose, other entry points can become unsupported/removed, which will enable further refactoring and tidying.

## It is a work in progress
This PR integrates the `testgen` and `testdriver` modules. Depending on feedback, I am planning to follow up with PRs to support the other modules.

### Added pyproject.toml
This new file captures pre-existing Python dependencies, which can now be installed with a single command - `pip install .`

### .gitignore'd DDT_DATA
I've assumed that the `/DDT_DATA` directory should be populated by the testgen script and therefore shouldn't contain any files by default. I've moved pre-existing contents to a new `./testData/examples/` directory and git-ignored the rest. This is to prevent people from committing testgen output by accident.


mradbourne:
Closing to prioritise other contributions

PR #331, opened 2024-10-29
Update NVM testing to be more robust

sven-oly:


echeran (approved):
LGTM so long as it works

PR #334, opened 2024-10-31
Update NodeJS generators to use .filter for subsampling

### LINKED ISSUE #283: Refactor JS test data generating fns to use `.filter()` when sampling
This issue could become obsolete if/when we generate data straight from CLDR. But for now, with the JS code, we can:

1. Refactoring test data sampling code to a common function rather than reimplement that logic in each component's test data generator
2. Instead of manual logic for taking every nth element of a collection, [reuse JS's FP functions and a one-liner](https://github.com/unicode-org/conformance/pull/280/files#diff-798cf7dceb35a95386801e14d3977bf28f437aab15f8e1c8e1e3ceb67cec65cdR50) ([alternative formulation](https://stackoverflow.com/a/41312896/2077918)).

sven-oly:
Resolves #283.

This makes sampling better for list_fmt and rdt_fmt test types when running subsets.


testgen/generators/common.js
```diff
@@ -7,10 +15,32 @@ function sample_tests(all_tests, run_limit) {
...
   let size_all = all_tests.length;
-  let increment = Math.floor(size_all / run_limit);
-  let samples = [];
-  for (let index = 0; index < size_all; index += increment) {
-    samples.push(all_tests[index]);
+  label_increment = Math.floor(size_all / run_limit);
+  max_count = run_limit;
+  added_count = 0;
+
+  if (debug) {
```
echeran, line 22: I filed https://github.com/unicode-org/conformance/issues/338 to capture the idea of improving & simplifying our JS logging code by using a library.

testgen/generators/common.js
```diff
@@ -7,10 +15,32 @@ function sample_tests(all_tests, run_limit) {
...
+    console.log('COMMON: ', size_all, ' increment: ', label_increment,
+                ' max_count: ', max_count);
+  }
+
+  let samples = all_tests.filter(filter_every_nth);
+  if (debug) {
+    console.debug('  COMMON results: ', samples.length);
+    console.debug('  COMMON results: ', samples);
```
echeran, line 30: I think this code would be cleaner if you just used `(i) => i % label_increment == 0` directly as the argument to `.filter()`

You don't even use `added_count` for debugging purposes because you can already get it from `samples.length`, which you do. Therefore, we can get rid of `filter_every_nth` altogether.

echeran (commented):
looks good modulo the simplification that is suggested

testgen/generators/common.js
sven-oly: Noted! What about https://github.com/pimterry/loglevel?

testgen/generators/common.js
sven-oly: Changed to (e, i) => i % label_increment == 0

where e is the element and i is the index.

testgen/generators/common.js
sven-oly: I removed the debugging code.

echeran (approved):

PR #335, opened 2024-11-01
Schema validations: reduce noise from unneeded logging messages

sven-oly:
This removes unnecessary logging that makes the log files harder to read.


echeran (approved):
Nice!

Issue #307, opened 2024-09-19
ICU4X and plural_rules: locale "root" gives an error.


sven-oly:

sven-oly:
This data results in an error with {"test_type": "plural_rules", "option":"root"}

{"test_type": "plural_rules", "locale":"root","label":"308","type":"cardinal","plural_type":"cardinal","sample":"15","hexhash":"3c749a7b7dfd26b68537283d16e03d72b6f869d4"}

Should these tests with locale "root" be dropped or mapped to "und" or something else?

sffc:
Please just remove locale "root" from the test data.

echeran:
Removing `root` as a locale from test data is suggested in #318

sven-oly:
This was fixedtoday  in PR#341.

Issue #317, opened 2024-09-30
Schema files: investigate ways to make these more consistent


sven-oly:
e.g., there are 3 schema files for each test type:
* test
* result
* verify
These all at least one common piece: "test_type" { "const", "datetime_fmt"}

And every test file should require "test_type" and "tests". Can we make a common skeleton or template for these schema files?

PR #341, opened 2024-11-12
Remove root locale from tests for locale names and plural rules

sven-oly:
Resolves issue #307


sven-oly:
This resolves issue #317

sffc (approved):

Issue #336, opened 2024-11-01
Reduce noise in output of scripts


sven-oly:
Debug logging produces some largely useless information, e.g., in Schema output.

Debug and other informational logging should be reduced, which will make it easier to find real problems in the end-to-end processing.

sffc:
@sffc to review and close if complete.

sven-oly:
Elango suggests that the logging to file is working correctly. Is there really too much output in the console? Is logging inconsistent?

sven-oly:
I think this is done. Let's close!

sffc:
The script still is a little noisy but it's probably fine.

PR #343, opened 2024-11-13
Reducing lots of noise from debug statements and unneeded detail in logs

sven-oly:
#336


sven-oly:
These changes aren't really the right method to reduce noise. Abandon?

echeran:
Yes, I propose to abandon since we have a file log handler set at `DEBUG` level and a console log handler set at `INFO` level. That already covers the original use case sufficiently well. FYI @sffc

Issue #309, opened 2024-09-23
Add CLDR 46 test data


sven-oly:

sven-oly:
Data generated from NodeJS must wait until Node has version with ICU-76.

sven-oly:
Note: the latest NodeJS release 23.2.0 still reports ICU version 75.1.

sven-oly:
ICU76.1 is supported in NodeJS 23.3.0.

sven-oly:
The full set of ICU76 test data is now in place. Resolving!

PR #348, opened 2024-11-27
Add NodeJS 23.3.0 supporting ICU76

sven-oly:
This also adds ICU76 testing for list format and relative date time because those generators use NodeJS data.


sven-oly:
Addresses #309

echeran (approved):
LGTM

PR #352, opened 2024-12-04
Better parsing of collationtest.txt data

sven-oly:
This fixes a lot of problems with parsing the Unicode test file collationtest.txt, handling the begin/end of regions of the data more carefully. This results in adding a few more tests (150 added in ICU75) as well as adding some attributes.

A new step is to use these attributes correctly in executors to compute the differences correctly.

### DIFF
--- a/schema/collation_short/test_schema.json
+++ b/schema/collation_short/test_schema.json
@@ -84,7 +84,7 @@
             "description": "indicates how case should be considered",
             "type": "string"
           },
-          "case_level": {
+          "caseLevel": {
             "description": "indicates the level  case should be considered",
             "type": "string"
           },
--- a/schema/collation_short/test_schema.json
+++ b/schema/collation_short/test_schema.json
@@ -48,6 +48,10 @@
             "description": "If collation should ignore punctuatio",
             "type": "boolean"
           },
+          "source_file" : {
+            "description": "source file of test data",
+            "type": "string"
+          },
           "line" : {
             "description": "line of the source of test data",
             "type": "integer"
@@ -60,8 +64,36 @@
             "description": "Strength of comparison: primary, secondary, ... identical",
             "type": "string"
           },
+          "alternate": {
+            "description": "shifted, non-ignorable",
+            "type": "string"
+          },
+          "numeric": {
+            "description": "sort as a number value",
+            "type": "string"
+          },
           "test_description": {
             "description": "Option field telling about this particular test",
+            "type": "string"
+          },
+          "reorder": {
+            "description": "instructions to reorder",
+            "type": "string"
+          },
+          "caseFirst": {
+            "description": "indicates how case should be considered",
+            "type": "string"
+          },
+          "case_level": {
+            "description": "indicates the level  case should be considered",
+            "type": "string"
+          },
+          "backwards": {
+            "description": "look at it backwards",
+            "type": "string"
+          },
+          "maxVariable": {
+            "description": "which variable is the maximum, e.g., currency",
             "type": "string"
           },
           "attributes": {
--- a/testgen/generators/base.py
+++ b/testgen/generators/base.py
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+# -*- coding: utf_8 -*-
 from abc import ABC, abstractmethod
 import copy
 import hashlib
@@ -69,7 +69,7 @@ class DataGenerator(ABC):
 
             # Create the 32 byte hasn, consisten with Javascript
             hasher = hashlib.sha1()
-            hasher.update(test_no_label_string.encode("utf-8"))
+            hasher.update(test_no_label_string.encode("utf_8"))
             hex_digest = hasher.hexdigest()
             test['hexhash'] = hex_digest
 
@@ -83,7 +83,7 @@ class DataGenerator(ABC):
                               filename)
 
         output_path = os.path.join(self.icu_version, filename)
-        output_file = open(output_path, "w", encoding="UTF-8")
+        output_file = open(output_path, "w", encoding="utf_8")
         json.dump(data, output_file, indent=indent)
         output_file.close()
 
@@ -133,7 +133,7 @@ class DataGenerator(ABC):
         if version:
             path = os.path.join(version, filename)
         try:
-            with open(path, "r", encoding="utf-8") as testdata:
+            with open(path, "r", encoding="utf_8") as testdata:
                 return json.load(testdata) if filetype == "json" else testdata.read()
         except BaseException as err:
             logging.warning("** readFile: %s", err)
--- a/testgen/generators/base.py
+++ b/testgen/generators/base.py
@@ -1,4 +1,4 @@
-# -*- coding: utf_8 -*-
+# -*- coding: utf-8 -*-
 from abc import ABC, abstractmethod
 import copy
 import hashlib
@@ -8,7 +8,6 @@ import logging.config
 import math
 import os
 import requests
-
 
 def remove_none(obj):
     # Recursively removes any parts with None as value
--- a/testgen/generators/collation_short.py
+++ b/testgen/generators/collation_short.py
@@ -205,10 +205,10 @@ class CollationShortGenerator(DataGenerator):
 
             # Detect comment at end of this rule line and keep it
             if rule_match := self.input_pattern_with_comment.match(line_in):
-                rule_list.append(rule_match.group(1).rstrip())
+                rule_list.append(rule_match.group(1).strip())
                 rule_comments.append(rule_match.group(2).strip())
 
-        rules = ' '.join(rule_list)
+        rules = ''.join(rule_list)  # Try combining without a space
         # If there's @rule, but no rules, we need to  indicate that rules should be reset.
         if rules == '':
             rules = ParseResults.RULE_RESET
--- a/testgen/generators/collation_short.py
+++ b/testgen/generators/collation_short.py
@@ -7,6 +7,20 @@ reblankline = re.compile("^\s*$")
 
 
 class CollationShortGenerator(DataGenerator):
+
+    def set_patterns(self):
+        self.root_locale = re.compile("@ root")
+        self.locale_string = re.compile("@ locale (\S+)")
+
+        self.test_line = re.compile("^\*\* test:(.*)")
+        self.rule_header_pattern = re.compile("^@ rules")
+        self.compare_pattern = re.compile("^\* compare(.*)")
+        self.comparison_line = re.compile("^([<=]\S*)(\s*)(\S*)(\s*)(.*)")
+
+        self.comment_pattern = re.compile("^(\S+)(\s*)#(.*)")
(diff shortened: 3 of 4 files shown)


testgen/generators/collation_short.py
```diff
@@ -78,8 +92,128 @@ def insert_collation_header(self, test_objs):
...
+            if self.root_locale.match(line_in):
+                break
+            if self.attribute_test.match(line_in):
+                break
+            if self.reorder_test.match(line_in):
+                break
+            if self.rule_header_pattern.match(line_in):
+                break
+            if self.test_line.match(line_in):
+                break
```
echeran, line 127: could you do?
```suggestion
            unexpected_line_patterns = [
                self.compare_pattern,
                self.local_string,
                ...
            ]
            if any([p.match(line_in) for p in unexpected_line_patters]):
                break
```

testgen/generators/collation_short.py
```diff
@@ -78,8 +92,128 @@ def insert_collation_header(self, test_objs):
...
+        # rule comments, and the new line index
+        # ignore comment at end of this rulrule_match.group(1)e line
+
+        rule_list = []
+        rule_comments = []
+
+        # Check if it's really the start of rules
+        line_in = lines[line_index]
+        if not self.rule_header_pattern.match(line_in):
+            return None, line_index
```
echeran, line 177: Ditto: assert rather than conditional

testgen/generators/collation_short.py
```diff
@@ -78,8 +92,128 @@ def insert_collation_header(self, test_objs):
...
+
+        line_index += 1
+        while line_index < len(lines):
+            line_in = lines[line_index]
+            if self.compare_pattern.match(line_in):
+                break
+            if self.locale_string.match(line_in):
+                break
+            if self.root_locale.match(line_in):
+                break
```
echeran, line 187: Ditto: similar FP style assertion / conditional?

testgen/generators/collation_short.py
echeran: Also: should we break out of the loop, or is it an error? Would an error signal an assumption about the test file that isn't true? And if the assumption about the test file isn't true, is that a problem that we need to balk at, or is it okay to keep going on?

testgen/generators/collation_short.py
```diff
@@ -78,8 +92,128 @@ def insert_collation_header(self, test_objs):
...
+
+            line_index +=1
+            if line_in == '' or line_in[0] == '#':
+                # Skip blank lines
+                continue
+
+            # It must be a rule line
+            # Remove any comments in the line preced&9<\x00                         # NUL not ignorableed by '#'
```
echeran, line 195: ??

testgen/generators/collation_short.py
```diff
@@ -98,52 +232,49 @@ def generateCollTestData2(self, filename, icu_version, start_count=0):
...
+            # line_number += 1
             line_in = raw_testdata_list[line_number]
-            line_number += 1
 
             is_comment = recommentline.match(line_in)
             if line_in[0:1] == "#" or is_comment or reblankline.match(line_in):
+                line_number += 1
                 continue
 
             if root_locale.match(line_in):
```
echeran, line 266: Can you add a line explaining that seeing a new locale means creating a new collator instance?

testgen/generators/collation_short.py
```diff
@@ -98,52 +232,49 @@ def generateCollTestData2(self, filename, icu_version, start_count=0):
...
+                attributes = {}
+
+                case_first = None
+                case_level = None
+                numeric = None
+                backwards = None
+                line_number += 1
                 continue
 
             locale_match = locale_string.match(line_in)
```
echeran, line 280: Can you add a line explaining that seeing a new locale means creating a new collator instance?

testgen/generators/collation_short.py
```diff
@@ -152,136 +283,110 @@ def generateCollTestData2(self, filename, icu_version, start_count=0):
...
+            if is_test_line:
+                # Get the description for subsequent tests
+                test_description = is_test_line.group(1)
+                line_number += 1
                 continue
 
             # Handle rules, to be applied in subsequent tests
-            is_rules = rule_header_pattern.match(line_in)
+            is_rules = self.rule_header_pattern.match(line_in)
             if is_rules:
```
echeran, line 306: Can you add a line explaining that seeing a new rule string means creating a new collator instance?

testgen/generators/collation_short.py
```diff
@@ -78,8 +92,128 @@ def insert_collation_header(self, test_objs):
...
+        locale = ""
+        strength = None
+
+    def parse_compare(self, line_index, lines):
+        # Handles lines in a compare region
+        # Test sections ("* compare") are terminated by
+        # definitions of new collators, changing attributes, or new test sections.
+        tests = []
+        line_in = lines[line_index]
+        if not self.compare_pattern.match(line_in):
```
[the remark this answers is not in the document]
sven-oly: Agreed. The problem would be for parse_compare (and also parse_rule) if the function is called when the current line is not "* compare" or "@ rule" as expected.

How about restructuring a bit to test the string within the function, proceeding only if the line content is as expected?

testgen/generators/collation_short.py
sven-oly: Yup, will do!

testgen/generators/collation_short.py
sven-oly: OK

testgen/generators/collation_short.py
sven-oly: Fixed

testgen/generators/collation_short.py
sven-oly: Changed the logic to use the functions to detect rules or comparisons. It not a match, return without advancing the line. If a match, then return the parsed results if t

testgen/generators/collation_short.py
```diff
@@ -92,47 +96,46 @@ def insert_collation_header(self, test_objs):
...
-            if self.rule_header_pattern.match(line_in):
-                break
-            if self.test_line.match(line_in):
+            # Time to end this set of comparison tests.
+            if any([p.match(line_in) for p in breakout_patterns]):
                 break
 
             # It's a blank line, comment, or a comparison
-            # ignore comment at end of this rulrule_match.group(1)e line
+            # ignore comment at end of this comment
```
echeran, line 132: This line doesn't say much. Remove? Reword?

testgen/generators/collation_short.py
```diff
@@ -92,47 +96,46 @@ def insert_collation_header(self, test_objs):
...
             if line_in == '' or line_in[0] == '#':
-                # Skip blank lines
+                # Skip blank and comment lines
                 line_index += 1
                 continue
 
+            # A comparison line starts with either "<" or "="
```
echeran, line 138: This comment belongs next to the regex because it explains why the regex is written the way it is. Copy-paste there. And optionally, reword this line to be more generic (ex: "check for a comparison"). That way, if the file format changes, and the regex pattern changes, you don't need to change this line b/c there's no duplication of content that leads to inconsistency & staleness.

testgen/generators/collation_short.py
```diff
@@ -141,73 +144,76 @@ def parse_compare(self, line_index, lines):
...
+                # ??? re-encode to get escaped version of s2?
                 new_test = {
                     'compare_type': compare_type,
                     's1': string1,
                     's2': string2,
                     'line': line_index,
                 }
-                if compare_comment:
+
+                if compare_comment:= is_comparison_match.group(5):
```
echeran, line 160: très cool

testgen/generators/collation_short.py
```diff
@@ -427,7 +400,7 @@ def generateCollTestDataObjects(
             line_number += 1
             if recommentline.match(item) or reblankline.match(item):
                 continue
-            # It's a data line.
+            # It's a data lin# Already joined!e.
```
echeran, line 403: Looks like one comment line teleported itself into another

verifier/testreport.py
```diff
@@ -641,7 +641,12 @@ def characterize_results_by_options(self, test_list, category):
                         'skeleton',
                         'language_label', 'locale_label',  # in lang_names
                         'option', 'locale',  # in likely_subtags
-                        'language_label', 'ignorePunctuation', 'compare_result', 'compare_type', 'test_description'
+                        'language_label', 'ignorePunctuation',
+                        # Some items for collation testing
```
echeran, line 645: move `'ignorePunctuation',` after this comment line b/c it is a Collator option

verifier/testreport.py
```diff
@@ -641,7 +641,12 @@ def characterize_results_by_options(self, test_list, category):
                         'skeleton',
                         'language_label', 'locale_label',  # in lang_names
                         'option', 'locale',  # in likely_subtags
-                        'language_label', 'ignorePunctuation', 'compare_result', 'compare_type', 'test_description'
+                        'language_label', 'ignorePunctuation',
```
echeran, line 644: how does `language_label` differ from `locale_label`?

schema/collation_short/test_schema.json
```diff
@@ -84,7 +84,7 @@
             "description": "indicates how case should be considered",
             "type": "string"
           },
-          "case_level": {
+          "caseLevel": {
```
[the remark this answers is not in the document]
sven-oly: I don't disagree, but the file collationtest.txt defines attributes with a variety of names including "caseLevel". Is it necessary to rename all these data types?

testgen/generators/collation_short.py
```diff
@@ -15,9 +19,9 @@ def set_patterns(self):
         self.test_line = re.compile("^\*\* test:(.*)")
         self.rule_header_pattern = re.compile("^@ rules")
         self.compare_pattern = re.compile("^\* compare(.*)")
-        self.comparison_line = re.compile("^([<=]\S*)(\s*)(\S*)(\s*)(.*)")
+        self.comparison_line = re.compile("^([<=]\S*)(\s*)(\S*)(\s*)#?(.*)")
 
-        self.comment_pattern = re.compile("^(\S+)(\s*)#(.*)")
+        self.rule_pattern_with_comment = re.compile("^([^#]+)#?(.*)")
```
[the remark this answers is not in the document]
sven-oly: OK

verifier/testreport.py
sven-oly: Tnx. See updated comments.

testgen/generators/base.py
```diff
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+# -*- coding: utf_8 -*-
```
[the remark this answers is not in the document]
sven-oly: OK, now its '_'

schema/collation_short/test_schema.json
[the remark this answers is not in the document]
echeran: Okay, it looks like there's a huge mix of camelCase and snake_case, so let's not worry about that here.

echeran (approved):
LGTM

PR #356, opened 2024-12-11
Update collation generator

sven-oly:
Also update characterizations for collation.

### DIFF
--- a/testgen/generators/collation_short.py
+++ b/testgen/generators/collation_short.py
@@ -205,10 +205,10 @@ class CollationShortGenerator(DataGenerator):
 
             # Detect comment at end of this rule line and keep it
             if rule_match := self.input_pattern_with_comment.match(line_in):
-                rule_list.append(rule_match.group(1).rstrip())
+                rule_list.append(rule_match.group(1).strip())
                 rule_comments.append(rule_match.group(2).strip())
 
-        rules = ' '.join(rule_list)
+        rules = ''.join(rule_list)  # Try combining without a space
         # If there's @rule, but no rules, we need to  indicate that rules should be reset.
         if rules == '':
             rules = ParseResults.RULE_RESET
--- a/testgen/generators/collation_short.py
+++ b/testgen/generators/collation_short.py
@@ -7,6 +7,20 @@ reblankline = re.compile("^\s*$")
 
 
 class CollationShortGenerator(DataGenerator):
+
+    def set_patterns(self):
+        self.root_locale = re.compile("@ root")
+        self.locale_string = re.compile("@ locale (\S+)")
+
+        self.test_line = re.compile("^\*\* test:(.*)")
+        self.rule_header_pattern = re.compile("^@ rules")
+        self.compare_pattern = re.compile("^\* compare(.*)")
+        self.comparison_line = re.compile("^([<=]\S*)(\s*)(\S*)(\s*)(.*)")
+
+        self.comment_pattern = re.compile("^(\S+)(\s*)#(.*)")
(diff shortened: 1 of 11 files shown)


verifier/testreport.py
```diff
@@ -651,15 +651,22 @@ def characterize_results_by_options(self, test_list, category):
                         'strength', 'caseFirst', 'backwards',
                         'reorder', 'maxVariable',
                         'source_file'
+                        # TODO!!! Characterize by actual_options keys & values
                         ]
             for key in key_list:
                 if test.get(key, None):  # For collation results
                     value = test[key]
+                    if not isinstance(value, str):
+                        # MAke this a string for comparison
```
echeran, line 660:
```suggestion
                        # Make this a string for comparison
```

verifier/testreport.py
```diff
@@ -651,15 +651,22 @@ def characterize_results_by_options(self, test_list, category):
...
+                        value = str(value)
                     if key not in results:
                         results[key] = {}
-                    if value not in results[key]:
-                        results[key][value] = set()
-                    results[key][value].add(label)
+                    try:
+                        if value not in results[key]:
+                            results[key][value] = set()
+                        results[key][value].add(label)
```
echeran, line 667: This can be simplified by only keeping the last line. This works because of the previous change you made to initialize `results` using `defaultdict` like so: `results = defaultdict(lambda : defaultdict(list))`. See code above for an example, [see here](https://realpython.com/python-defaultdict/#understanding-the-python-defaultdict-type) for an explanation.
```suggestion
                        results[key][value].add(label)
```

verifier/testreport.py
```diff
@@ -748,24 +758,28 @@ def characterize_datetime_tests(self, test_list, results):
...
+                            else:
+                                results['input_size'].add(len(value))
+                        if key == 'rules':
+                            value = 'RULE'  # A special case to avoid over-characterization
+                        if key not in results:
+                            results[key] = {}
+                        try:
+                            if not results[key].get(value, None):
+                                results[key][value] = set()
+                            results[key][value].add(label)
```
echeran, line 776: See if you can assure that wherever the initialization of `input_data` occurs uses `defaultdict`, and then simplify the code here.

testgen/generators/collation_short.py
```diff
@@ -142,7 +142,7 @@ def check_parse_compare(self, line_index, lines):
                 raw_string2 = is_comparison_match.group(3)
                 string2 = ''
                 try:
-                    string2 = raw_string2.encode().decode("unicode_escape")
+                    string2 = raw_string2  # Don't do any unescaping
```
echeran, line 145: Hmm, if we're not unescaping the escape sequences, then how are we comparing the intended characters in the string?...

echeran (changes requested):
please fix

verifier/testreport.py
sven-oly: OK

verifier/testreport.py
sven-oly: Done

verifier/testreport.py
sven-oly: Trying this now.

testgen/generators/collation_short.py
sven-oly: I will look at handling these in the test **driver**, making sure that we do indeed send the right characters to the executors.

echeran (approved):
LGTM, filing https://github.com/unicode-org/conformance/issues/358 to capture the unaddressed comments for later follow up

PR #360, opened 2024-12-17
Refactor Rust code and add 1.5

sffc:
Fixes https://github.com/unicode-org/conformance/issues/302
Replaces https://github.com/unicode-org/conformance/pull/239


sven-oly (approved):
Yay!

sven-oly (approved):

PR #366, opened 2024-12-19
ICU4X 2.0 initial datetime support

### LINKED ISSUE #355: ICU4X - need version 2.0 with ICU 76 / CLDR 46 data

sffc:
Should be enough to fix #355

There are some failures and errors that need to be investigated. But the tests are running and 4000 of them are passing.

image


sven-oly (approved):
But please fix the conflict.

sven-oly:
I see one conflict, but all else looks OK. I'll approve. Good work!

PR #369, opened 2024-12-19
Reduce debug noise for issue #336

sven-oly:
This removes excessive debug output from conformance test logs.


schema/check_generated_data.py
```diff
@@ -15,20 +15,20 @@
 import schema_validator
 from schema_files import ALL_TEST_TYPES
 
+logger = logging.Logger("Checking Test Data vs. Schemas LOGGER")
+logger.setLevel(logging.INFO)
```
sffc, line 19: Suggestion: often these libraries have a way to set the logging level from an environment variable or from the CLI, so that you can do something like `DEBUG=* ./genData` or `./genData -vv`

sven-oly: Yes, that's possible. I'll take a look.

Also, I've removed lots of excessive logging in the latest commits.

One more thing: when executing in parallel, the individual threads cannot have a logger in their objects because a logger "cannot be pickled". That's a small complication / limit to logging.

sffc: I think don't pickle the logger; just make a new one for each thread

schema/schema_validator.py
```diff
@@ -449,7 +447,7 @@ def main(args):
     base_folders, test_types, result_folders = process_args(args)
 
     logger = logging.Logger("TEST_GENERATE LOGGER")
-    logger.setLevel(logging.INFO)
+    logger.setLevel(logging.WARNING)
```
echeran, line 450: By hard-coding the logger level for this file to be WARNING, and given that most of the logging statements here are either DEBUG or INFO, you're effectively turning off all of the logging happening in this module. Is that your intention?

testdriver/testdriver.py
```diff
@@ -18,11 +18,12 @@
 
 from testplan import TestPlan
 
+logger = logging.Logger("TEST DRIVER LOGGER")
+logger.setLevel(logging.WARNING)
```
echeran, line 22: Unless there's a good reason, we should not manually (hardcode) the logger level in code. Instead, we should allow the logger level to be set by our existing config file. So here & other places in the PR, where you see `logger.setLevel(...)`, we should remove those statements.

testgen/generators/datetime_fmt.py
```diff
@@ -156,15 +156,8 @@ def process_test_data(self):
             result = self.generate_datetime_data_from_cldr(dt_json_path, self.run_limit)
             return result
 
-        # OK, there's no CLDR-based JSON data available.
-        run_list = [
```
sffc, line 160: Question: is this dead code?

testgen/generators/list_fmt.py
```diff
@@ -27,12 +27,6 @@ def process_test_data(self):
             exec_list.append('-run_limit')
             exec_list.append(str(self.run_limit))
 
-        run_list = [
```
sffc, line 30: Question: Is this dead code?

sffc (commented):
Seems fine

Issue #205, opened 2024-03-25
Error handling for expected likely subtag errors


echeran:
Some of the test cases for likely subtags are supposed to fail. The source test data encodes this information.

We need to decide a canonical error representation during our testing. [CLDR defines  ways to handle such an error](https://unicode.org/reports/tr35/#Likely_Subtags):

```
Signalling an error can be done in various ways, depending on the most consistent approach for APIs in the module. For example:

raise an exception
return an error value (such as null)
return the input (with missing fields)
return the input, but "Zzzz", and/or "ZZ" substituted for empty fields.
"und"
```

Whichever way we decide to handle it, we need to update our executors, which will also hopefully decrease the number of test failures.

sffc:
For likely subtags, ICU4X has this enum:

```rust
pub enum TransformResult {
    Modified,
    Unmodified,
}
```

Let's supposed that enum has a third variant: `NoData`

Code to create a handled exception would look something like this:

```rust
let mut locale = // get from input
let result = icu4x_locale_expander.maximize(&mut locale);
if matches!(result, TransformResult::NoData) {
  // return handled exception
} else {
  // return a completed result
}
```

sffc:
Also see my comment in https://github.com/unicode-org/conformance/issues/242

I think there are 3 ways to solve this issue:

1. Add code to the executors to detect when a minimize/maximize operation failed: for example, if the output is unmodified and there were fewer than 3 subtags in a maximize operation. In those cases, the executor should return some sort of failure status, and the verifier should detect that status and return a successful status.
2. Consider it a Known Issue when an executor does not have the capability to distinguish a likely subtags operation that did not run to completion, and if an implementation adds such an API (as I sketched out above in ICU4X) it can start passing those tests
3. Change the test expectations to be input == output, removing the "expected exception" case entirely. Implementations that _do_ handle the exception should return input as output.

sven-oly:
New observatio: Most of the test failures on likely subtags are for the "favorScript" option.

However, the APIs do not allow minimizing on either script or region.

sffc:
These tests are testing the following language in UTS 35:

> Signalling an error can be done in various ways, depending on the most consistent approach for APIs in the module. For example:
>
> 1. raise an exception
> 2. return an error value (such as null)
> 3. return the input (with missing fields)
> 4. return the input, but "Zzzz", and/or "ZZ" substituted for empty fields.
> 5. "und"

[...]

PR #371, opened 2025-01-02
LikelySubtags: - minimizeFavorScript --> unsupported.  Fixes many errors in ICU4C, ICU4J, and Node

sven-oly:
Reclassifying any with favor script to unsupported. Also setting result for all "qaa" locale items to "FAIL" because these are note valid language tags.

Note that there are still errors reported in #205 that need to be examined.

Also, ICU4X still has issues.


sven-oly:
#205

sven-oly:
Note that this now fixes all likely subtags in ICU4X 2.0, and almost all in ICU4X 1.3 - 1.5.

Please take a look!

executors/cpp/likely_subtags.cpp
```diff
@@ -50,6 +50,19 @@ string TestLikelySubtags(json_object *json_in) {
   string name_string;
   StringByteSink<string> byteSink(&name_string);
 
+  // Check for special case of reserved codes for local use
+  const char* lang_code = displayLocale.getLanguage();
+  const string lang_code_string(lang_code);
+  if (lang_code_string >= "qaa" && lang_code_string <= "qtz") {
```
sffc, line 56: Issue: It's not correct for the executor to specifically sniff for qaa. There are many subtags that could cause this type of behavior, not just qaa. The executor should sniff for some other signal of a failure status

testgen/icu76/likelySubtags.txt
```diff
@@ -1441,7 +1441,7 @@ und-Latn-MQ ;	fr-Latn-MQ ;	fr-MQ ;
 und-Latn-MR ;	fr-Latn-MR ;	fr-MR ;	
 und-Latn-MS ;	en-Latn-MS ;	en-MS ;	
 und-Latn-MT ;	mt-Latn-MT ;	mt ;	
-und-Latn-MU ;	en-Latn-MU ;	en-MU ;	
+und-Latn-MU ;	mfe-Latn-MU ;	mfe ;
```
sffc, line 1444: Question: Why did this data change?

echeran:
If `qaa` and other language codes are reserved, then we should consider whether they are valid test data to begin with. Given that the original source file with test data coming from CLDR (via ICU) just says `FAIL` in the place of the expected value  when the input contains `qaa`, it's a sign that an error is expected. And I think it doesn't make sense to expect real world data about scripts and regions to interact with a _reserved_ language code, we should remove that from the test data, I think.

sven-oly:
OK, I agree that we could/should removed the non-legal language codes in the test generator.

executors/cpp/likely_subtags.cpp
sven-oly: OK, I'll change the test generator to remove this data range.

sven-oly:
Another question about this: should we mark minimize preferring script as  "unsupported" for C and J?

There's no public API way to support this, although ICU4X does indeed allow preferring the script or the region.

sven-oly:
Note the qaa codes just generate the same data as the input. This is one way to fix this problem.

executors/icu4j/74/executor-icu4j/src/main/java/org/unicode/conformance/testtype/likelysubtags/LikelySubtagsTester.java
```diff
@@ -37,8 +37,9 @@ public ITestTypeOutputJson execute(ITestTypeInputJson inputJson) {
     try {
       output.result = getLikelySubtagString(input);
     } catch (Exception e) {
-      output.error = e.getMessage();
-      output.error_message = e.getMessage();
+      output.error_type = "unsupported";
+      output.unsupported = e.getMessage();
+      output.error_detail = e.getMessage();
```
sffc, line 42: Issue: don't mark a test as unsupported without constraining the error type

sven-oly: Sure. I've updated the catch to be specific.

sven-oly:
PTAL

sffc (approved):

PR #374, opened 2025-01-03
CPP files updated with clang

sven-oly:
Added so 'make clang' runs, updating files as suggested.

Updates .cpp files, too.

### DIFF
--- a/executors/cpp/list_fmt.cpp
+++ b/executors/cpp/list_fmt.cpp
@@ -29,7 +29,7 @@ using std::endl;
 using std::string;
 
 /* Main test function */
-string TestListFmt (json_object* json_in) {
+auto TestListFmt (json_object* json_in) -> string {
   UErrorCode status = U_ZERO_ERROR;
 
   json_object* label_obj = json_object_object_get(json_in, "label");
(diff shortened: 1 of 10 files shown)


sven-oly:
Updates the CPP code with more modern options. PTAL

executors/cpp/list_fmt.cpp
```diff
@@ -113,7 +113,7 @@ string TestListFmt (json_object* json_in) {
     return json_object_to_json_string(return_json);
   }
 
-  UnicodeString *u_array = &u_strings[0];
+  UnicodeString *u_array = u_strings.data();
```
echeran, line 116: This is a strange suggestion from Clang. I think you want [`vector::front`](https://cplusplus.com/reference/vector/vector/front/)?

echeran (commented):
One suggestion. Otherwise looks good.

executors/cpp/list_fmt.cpp
sven-oly: No, .front() gives the first element's value, not its location. In this case .data() effectively gives a pointer to the start of the array.

sven-oly: https://cplusplus.com/reference/vector/vector/data/

echeran: Okay, is see the difference in the data type of the returned value between the two functions. And yeah, `.data()` returns a pointer to the internal array, and C++ guarantees the array of a vector is entirely contiguous in memory, so this works.

echeran (approved):
LGTM

PR #376, opened 2025-01-06
Removing old target DDT_DATA

### LINKED ISSUE #180: Remove `DDT_DATA` dir and scripts referencing it
The `DDT_DATA` directory is obsolete at this point, and it seems to be just a copy of a portion of the `TEMP_DATA` directory that get created locally to store intermediate files.

We should remove the `DDT_DATA` directory. At this point, all scripts referencing that directory are obsolete, too.

Do not remove any Python code references to `ddt_data`. The Python identifier is the alias used for `datasets.py` when importing that Python file/module.

sven-oly:
Fixes #180


echeran (approved):
Yay!

Issue #186, opened 2024-02-29
Test type "collation_short" should be renamed to "collation" throughout the project


sven-oly:

sven-oly:
In progress...

PR #383, opened 2025-01-10
Change 'collation_short' to 'collation' everywhere

sven-oly:
This is an old issue, simply to rename this component. This does not change anything else.


sven-oly:
#186

sven-oly:
Merging this because it may block other progress.

PR #384, opened 2025-01-10
Add sys.exit(1) on failures

sven-oly:


schema/check_schemas.py
```diff
@@ -126,7 +126,8 @@ def main(args):
                       len(schema_errors), schema_count)
         for failure in schema_errors:
             logging.error('  %s', failure)
-        exit(1)
+        # We need to clobber the process
+        sys.exit(1)
     else:
         logging.info("All %d schema are valid in file %s", schema_count, output_filename)
         exit(0)
```
echeran, line 133: You didn't change this line. But I wonder if you need it.

schema/check_generated_data.py
```diff
@@ -89,17 +89,17 @@ def main(args):
...
         logging.critical('Test data file files: %d fail out of %d:',
                          len(schema_errors), schema_count)
         for failure in schema_errors:
             logging.critical('  %s', failure)
-        exit(1)
+        sys.exit(1)
     else:
         logging.info("All %d generated test data files match with schema", schema_count)
-        exit(0)
+        sys.exit(0)
```
echeran, line 102: Do you need a `sys.exit(0)`?  If this is the last line of the happy case, then it reads better to remove the line. When I see `sys.exit(<nonzero>)`, I assume that someone is signalling an _unnaturally early_ exit, so a really unhappy case. Happy cases and handled unhappy cases would otherwise be treated "normally" (no op, try/catch exception handling, etc.).

schema/check_schemas.py
sven-oly: I think sys.exit is appropriate here.

schema/check_generated_data.py
sven-oly: Fixed.

echeran (approved):

PR #385, opened 2025-01-11
Add more dart versions: 74, 75, and 76

sven-oly:
This seems to all that's necessary to bring dart_web up to date with ICU.


sven-oly:
This brings the ICU version support in Dart Web up to date by adding ICU74, ICU75, and ICU76.

It also makes #217 obsolete.

echeran (approved):
LGTM

PR #387, opened 2025-01-14
Fix data generation from nodejs with `nvm run` plus command

sven-oly:


echeran (approved):
LGTM. I ran locally and confirm the desired behavior.

Issue #388, opened 2025-01-14
DartWeb executor prints voluminous messages for all returns of "error" or "unsupported"


sven-oly:
DartWeb prints out an error message for every test result that contains "error". However, this is excessive, clogging up the debug.log files and causing problems with continuous integration. This was exacerbated by adding ICU versions 74, 75, and 76 to run_config.json in #385

echeran:
In addition, I wonder if Dart Web is causing the "hanging" behavior that we are seeing in the pipeline end-to-end runs because Dart Web executor code is not doing a "system.exit()" equivalent when it encounters an error. Ending the process upon an uncaught exception/error is the proper behavior.

Maybe updating Dart Web code to terminate upon uncaught errors/exceptions would automatically allow the Python thread to terminate so that the Python thread pool knows to terminate and thus propagate the error all the way up to the CI so that the job fails.

PR #390, opened 2025-01-14
Dart web: remove extraneous debug output

sven-oly:
Printing out every case where "error" is in the output is unnecessary and produces huge numbers of log lines.


sven-oly:
#388

echeran (approved):
LGTM. Thank you, hoping this will clean up the output

Issue #381, opened 2025-01-08
Schema failures on message format 2 in iCU76


sven-oly:

sven-oly:
@mradbourne
I see this error message in icu76 generated test files for message_fmt2.

```
  File "/usr/local/google/home/ccornelius/ICU_conformance/conformance/testgen/generators/message_fmt2.py", line 54, in process_test_data
    logging.error("Problem validating JSON: %s agains schema",
Message: 'Problem validating JSON: %s agains schema'
Arguments: ('/usr/local/google/home/ccornelius/ICU_conformance/conformance/testgen/generators/../icu76/message_fmt2/message-format-wg-tests/functions/datetime.json', PosixPath('/usr/local/google/home/ccornelius/ICU_conformance/conformance/testgen/generators/../../schema/message_fmt2/testgen_schema.json'))
[2025-01-09 15:53:02,778] [ERROR]  Additional properties are not allowed ('$schema' was unexpected)
Failed validating 'additionalProperties' in schema:
    {'$comment': 'This allOf specifies required test properties that allow '
                 'a default. A value will be required in '3. 

                 "'defaultTestProperties' if one is not provided for every "
                 'individual test.',
     '$defs': {'anyExp': {'anyOf': [{'required': ['exp']},
                                    {'required': ['expParts']},
                                    {'required': ['expErrors']}]},
               'bidiIsolation': {'description': 'The bidi isolation '
...
     '$id': 'https://raw.githubusercontent.com/unicode-org/message-format-wg/main/test/schemas/v0/tests.schema.json',
     '$schema': 'https://json-schema.org/draft/2020-12/schema',
     'additionalProperties': False,
...
     'required': ['tests'],
     'title': 'MessageFormat 2 data-driven tests',
     'type': 'object'}
On instance:
    {'$schema': 'https://raw.githubusercontent.com/unicode-org/message-format-wg/main/test/schemas/v0/tests.schema.json',
     'defaultTestProperties': {'expErrors': False, 'locale': 'en-US'},
     'description': 'The built-in formatter for datetimes.',
     'scenario': 'Datetime function',
     'tests': [{'exp': '{:datetime}',3. 

                'expErrors': [{'type': 'bad-operand'}],
                'src': '{:datetime}'},3. 

...
```

This is one of the generated test files that includes "$schema" property, but the file testgen_schema.json doesn't allow that as a property. I think all of them are actually failing schema validation

Elango and I see two options:
1. Allow additional properties
[...]

PR #394, opened 2025-01-16
Resolves issues in ICU75 and ICU76 schema checks for MF2

sven-oly:
ICU75 has a field expCleanSrc that was removed in ICU76.

And ICU76 includes "$schema" as source data. Removing this fixes many schema failure messages.


sven-oly:
#381

sven-oly:
Please take a look.

sven-oly:
Please take a look. The schemas are failing with the ICU76 MF2 test data!

echeran (approved):
LGTM. This works around what seems to me to be a deficiency in the Python JSON Schema library. The `$schema` property name is valid JSON Schema according to the spec, so it should not be causing the problems that @sven-oly has been experiencing seemingly spurious error messages.

PR #396, opened 2025-01-25
Advanced collation - giving more info on CPP failures

sven-oly:
This puts the actual information received into the output so we can debug.


executors/cpp/coll.cpp
```diff
@@ -217,9 +222,13 @@ auto TestCollator(json_object *json_in) -> string {
...
     // Default checking for <= 0.
     coll_result = (uni_result != UCOL_GREATER);
-  } else {
-    coll_result = (uni_result == UCOL_EQUAL);
-  }
+  } else
+    if (compare_type_string == "=") {
+      coll_result = (uni_result == UCOL_EQUAL);
+    } else {
+      //
```
echeran, line 229: Did you forget to fill in a comment here? I assume that we test against UCOL_EQUAL as a default because we don't have any other compare types except `=` and the implicit `<=`, and `=` is a fine enough default. Is that correct?

echeran (approved):
LGTM, but see comments.

PR #402, opened 2025-03-04
Small fixes to Dart

mosuem:
Fix implicit call to `toJson` and run `dart pub upgrade --major-versions` for `dart_native` and `dart_web`


sven-oly (commented):
I think this looks OK, but I'm having trouble running the bash jobs locally, with*all* Dart tests reporting "Error" or "Not supported".

sven-oly (approved):
OK, I'll approve. Let's see what happens.

PR #403, opened 2025-03-10
Add known issues

sven-oly:
Recognizing some types of test failures as Known Issues:

1. Likely Subtags: und-Latn-RS should resolve to sr-Latn and sr-Latin-RS.

2. Language Names: several locale options such as -un-ms and "fonipa" are not handled correctly. These are known issues.

3. Language Names: using parentheses rather than brackets in strings.
4.


verifier/check_known_issues.py
```diff
@@ -184,6 +191,75 @@ def check_rdt_known_issues(test):
...
+             langname_brackets(test))
+    return is_ki
+
+
+def langname_fonipa(test):
+    # Fonipa - one of the less supported locale options.
+    input_data = test['input_data']
+    lang_label = input_data['language_label']
+    if lang_label.find('fonipa') >= 0:
+        return knownIssueType.langnames_fonipa
```
sffc, line 235: Discussion: fonipa might be a correlation but not the actual bug

sven-oly:
Please take a look. We need to post ICU4C/J issues for some of these

verifier/check_known_issues.py
sffc: Suggestion: be a bit more precise: check `scouse-fonipa`

sffc (approved):

sven-oly:
Note that previous DateTime known issue was too broad with the "known_issue_id.Alternate formatting with "at" between time and date" detection. This PR fixes that, and also adds checks for:
1.  Arabic vs. ASCII comma
2. Insertion of a comma in some date/time formatted results.

PR #407, opened 2025-03-18
Highlight no fail or error

sven-oly:
This adds color to the summary page links for each component and ICU version. It also provides hover text describing the label.

Also, the labels on the colored bars for resutls are simplified if the platform version contains the ICU version, e.g., "73.2, 73" --> "73.2".


sffc (approved):
#aa8400 is such an ugly color :)

sven-oly:
I'm open to a different color for no failures or errors. It's not quite green but partly orange and gray.

PR #409, opened 2025-03-18
Add ICU4X beta2

sffc:


sven-oly (approved):

sffc:
Note: this may cause beta1 to disappear from the dashboard due to a bug (both are currently using 76 data). This should be fixed when 77 gets added and we can make beta2 use 77 data.

PR #419, opened 2025-03-25
Skip unsupported locales in plural rules test in ICU4X

sffc:
Improves ICU4X plurals conformance.

However, I think we should just not emit the non-modern locales in test data gen.


sven-oly (approved):
Thanks!

echeran (approved):
LGTM

sffc:
Ok, I'll merge this for now, and I opened #422 to follow-up with the better solution later.

PR #421, opened 2025-03-25
Plural rules known issue

sven-oly:
Classify test failures with non-integral plural samples as "known issues". This fixes many test failures in ICU4C, ICU4J, and ICU4X.

### DIFF
--- a/executors/node/plural_rules.js
+++ b/executors/node/plural_rules.js
@@ -28,8 +28,13 @@ module.exports = {
                                 }
                };
       }
-      if (raw_sample.indexOf('.') >= 0) {
+      let index_of_decimal = raw_sample.indexOf('.');
+      if (index_of_decimal >= 0) {
         sample = parseFloat(raw_sample);
+        // Workaround for issue : https://github.com/tc39/ecma402/issues/397
+        let num_fraction_digits = raw_sample.length - index_of_decimal -1;
+        test_options['minimumFractionDigits'] = num_fraction_digits;
+        test_options['maximumFractionDigits'] = num_fraction_digits;
       } else {
         sample = parseInt(raw_sample);
       }
@@ -75,7 +80,7 @@ module.exports = {
 
     let list_formatter;
     try {
-      p_rules = new Intl.PluralRules(actual_locale, test_options);
+     p_rules = new Intl.PluralRules(actual_locale, test_options);
     } catch (error) {
       /* Something is wrong with the constructor */
       return_json['error'] = 'CONSTRUCTOR: ' + error.message;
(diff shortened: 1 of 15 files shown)


verifier/check_known_issues.py
```diff
@@ -287,6 +290,18 @@ def langname_brackets(test):
         return None
 
 
+def check_plural_rules_issues(test):
+    try:
+        input_data = test['input_data']
+        sample_string = input_data['sample']
+        # Plural rules for floating point values may not be supported
+        if sample_string.find('.') >= 0:
+            return knownIssueType.plural_rules_floating_point_sample
```
sffc, line 299: You should also check that the final digit is `0`

executors/node/plural_rules.js
```diff
@@ -28,8 +28,13 @@ module.exports = {
...
                };
       }
-      if (raw_sample.indexOf('.') >= 0) {
+      let index_of_decimal = raw_sample.indexOf('.');
+      if (index_of_decimal >= 0) {
         sample = parseFloat(raw_sample);
+        // Workaround for issue : https://github.com/tc39/ecma402/issues/397
+        let num_fraction_digits = raw_sample.length - index_of_decimal -1;
+        test_options['minimumFractionDigits'] = num_fraction_digits;
+        test_options['maximumFractionDigits'] = num_fraction_digits;
```
sffc, line 37: Unrelated diff

sffc (changes requested):

verifier/check_known_issues.py
sven-oly: Probably check if all digits after "." are "0", not just the last one.

sven-oly: Also changed so  test reported as "unsupported" by the executor are not reassigned as "known issues".

sven-oly: Done

verifier/check_known_issues.py
```diff
@@ -37,6 +38,8 @@
 NBSP = '\u202f'
 SP = '\u0020'
 
+floating_point_all_zero = re.compile(r"\.[^1-9]+")
```
sffc, line 41: We only care if the last digit after the decimal is zero. I think a better regex would be:

```suggestion
floating_point_all_zero = re.compile(r"\.\d*0$")
```

verifier/check_known_issues.py
```diff
@@ -37,6 +38,8 @@
 NBSP = '\u202f'
 SP = '\u0020'
 
+floating_point_all_zero = re.compile(r"\.\d*0$")
```
sffc, line 41:
```suggestion
floating_point_has_trailing_zero = re.compile(r"\.\d*0$")
```

sffc (approved):

sffc (approved):

PR #428, opened 2025-03-31
Plural rules ICU4J handle known issue with '4.1'

sven-oly:
This should be solved with better handling of precision, but there's no public API.


sven-oly:
Will fix #421

sffc (changes requested):
This is not a good test; 4.1 is a different problem than the trailing zeros

sffc (approved):

sven-oly:
There's a new KI for this.

PR #436, opened 2025-04-09
CLDR 47 date time - add skeleton to test data

sven-oly:
Updates data generated to set skeleton and semantic skeleton fields.

Also characterize differences in AM/PM, i.e., AM vs. PM or case, etc.


verifier/testreport.py
```diff
@@ -751,6 +751,8 @@ def characterize_plural_rules_tests(self, test_list, results):
 
     def characterize_datetime_tests(self, test_list, results):
         # look for consistencies with datetime_fmt test
+        if not test_list:
```
echeran, line 754: suggestion: file an issue for the root cause, note the issue number in a comment here

This workaround is needed in order to make the end-to-end process complete. It would otherwise fail due to badly formed data caused by upstream bugs in one of the executors. However, this workaround needs to be removed ASAP because it will otherwise silently remove those error-inducing test cases from our output reporting as if they never existed for the culprit executors. That would effectively be a silent fail, which is a hard thing to detect and fix if it were left to the future.

We should immediately fix the underlying cause, so having the issue helps address that. Then the comment here allows us to know to remove this workaround at the same time.

sven-oly: I agree with your comments. I've removed this test, which was a hack.

sven-oly: I've also added characterizations that find some common problems with AM/PM in test results.

sven-oly:
Note that there are some strange test failures, e.g., AM vs PM. Also, non-breaking spaces vs. ASCII space in C++ results.

echeran (approved):
LGTM

PR #443, opened 2025-04-23
Improve ICU4X beta2 datetime executor

sffc:
Not yet completely done, but it brings the number of errors from ~90 down to 12 and successes from 0 to 25.

### DIFF
--- a/executors/rust/src/datetime_2_beta2.rs
+++ b/executors/rust/src/datetime_2_beta2.rs
@@ -8,6 +8,7 @@ use icu::datetime::input::ZonedDateTime;
 use icu::datetime::options::*;
 use icu::datetime::DateTimeFormatter;
 use icu::datetime::DateTimeFormatterPreferences;
+use icu::time::zone::UtcOffsetCalculator;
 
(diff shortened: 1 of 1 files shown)


sffc:
Also, the input options are something like:

```
{"locale":"en","input_string":"2000-01-01T00:00:00Z","options":{"timeZone":"Etc/GMT","dateStyle":"short","timeStyle":"full","calendar":"gregory"},"tz_offset_secs":0,"label":"012","original_input":"2000-01-01T00:00Z[Etc/GMT]","dateTimeFormatType":"atTime","hexhash":"7ab8fdbd6793958b9b55a0c11be82098cd3d7581"}
```

A few issues with this:

1. Doesn't include fields such as time precision, zone style, year style, and alignment
2. It seems arbitrary whether things are in the `options` or not

sven-oly:
> 1. Doesn't include fields such as time precision, zone style, year style, and alignment

- zoneStyle and yearStyle are in the data from CLDR. 
- time precision and alignment are not provided now.

PR #450, opened 2025-05-03
Update names of files for "lang_names" to be consistent across phases of testing

sven-oly:
Several strings were used for file names with this test type. Making it consistent will support analysis of data, making it easier to compare data with the same test types across platforms, ICU versions, and steps of the test process (testgen, testdriver, schema analysis, and verification.)


sven-oly:
#450

PR #456, opened 2025-05-08
Use stable version of Dart in CI

echeran:
Lots of CI end-to-end workflow instances are failing in PRs and on `main`. The errors in the workflow logs only pertain to Dart, and they indicate a version that is newer. A local run of the end-to-end script succeeds successfully, and a stable version of Dart is being used.

This will turn off using unstable versions of Dart so that we still to stable ones. 

Hopefully, this will fix the noisy CI errors that are getting in the way of unrelated PRs.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -32,8 +32,6 @@ jobs:
 
       - name: Setup version of Dart
         uses: dart-lang/setup-dart@d6a63dab3335f427404425de0fbfed4686d93c4f
-        with:
-          sdk: dev
       - name: Set version of Rust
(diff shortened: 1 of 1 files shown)


echeran:
@mosuem All of @sven-oly's pending PRs are blocked due to CI errors that all have error lines that have something to do with Dart. We don't see this locally. We noticed that CI is using the dev channel for Dart. What was the original reason for that? Can we use stable instead? Also, our dashboard is full of lots of errors for Dart, and it's too many to leave there without fixing. Otherwise, it looks like we're just a canary system for experimental versions of a library, when we actually need to test major versions intended for users.

sven-oly (approved):

PR #458, opened 2025-05-15
Add segmenter as test type for nodejs with minimal data

sven-oly:


schema/segmenter/test_schema.json
```diff
@@ -0,0 +1,110 @@
...
+ "title": "ICU Conformance test data description",
+ "description": "This documents the format of segmenter test data for execution",
+ "type": "object",
+ "properties": {
+   "additionalProperties": false,
+   "test_type": {
+     "description": "The name of the test",
+     "const": "segmenter"
+   },
+   "Test scenario": {
```
echeran, line 12: check identifier (has a space and initial capital letter)

remove if not used

schema/segmenter/test_schema.json
```diff
@@ -0,0 +1,110 @@
...
+   },
+   "source_file": {
+     "description": "origin of conformance data",
+     "type": "string"
+   },
+   "source_version": {
+     "description": "version of conformance data",
+     "type": "string"
+   },
+   "test_ environmment": {
```
echeran, line 24: check identifier (has a space)

are these being used? if not, then remove

schema/segmenter/test_schema.json
```diff
@@ -0,0 +1,110 @@
...
+           "type": "string"
+         },
+         "options": {
+           "description": "Formatting paramters for the number",
+           "type": "object",
+           "properties": {
+             "granularity": {
+               "type": "string",
+               "description": "What kind of segmentation to do",
+               "enum": ["grapheme", "word", "sentence"]
```
echeran, line 96: I prefer `grapheme_cluster` over `grapheme` because that is more precise. `grapheme` is confusing, even if it comes from ECMA.

schema/segmenter/result_schema.json
```diff
@@ -0,0 +1,67 @@
+{"$schema": "https://json-schema.org/draft/2020-12/schema",
+ "$id": "https://github.com/unicode/conformance/segmenter_verify_schema.json'",
+ "title": "ICU Plural Rules verify data description",
```
echeran, line 3: modify

testdriver/datasets.py
```diff
@@ -195,6 +196,12 @@ def def_value():
                                  'plural_rules_verify.json',
                                  CLDRVersion.CLDR44, ICUVersion.ICU74)
 
+testName = 'segmenter'
+testDatasets[testName] = DataSet(testType.rdt_fmt.value,
+                                 'segmenter_test.json',
+                                 'segmenter_verify.json',
+                                 CLDRVersion.CLDR44, ICUVersion.ICU74)
```
echeran, line 203: fix version nums

verifier/testreport.py
```diff
@@ -820,6 +823,9 @@ def check_simple_text_diffs(self, test_list, category):
             label = fail['label']
             actual = fail.get('result', None)
             expected = fail.get('expected', None)
+            if type(actual) != type(expected):
```
echeran, line 826: I don't think this should be necessary because the actual & expected for the test cases we run should at least have the same type. If not, then our test case generation is severely wrong somehow.

However, we shouldn't gloss over that by ignoring the problem, so we should get rid of this `continue` here.

verifier/testreport.py
```diff
@@ -921,6 +927,28 @@ def check_simple_text_diffs(self, test_list, category):
...
+        for check in all_checks:
+            results[check] = set()
+
+        for fail in test_list:
+            label = fail['label']
+            actual = fail['result']
+            expected = fail.get('expected', None)
+            if type(actual) != type(expected):
+                # This is a type mismatch
+                continue
```
echeran, line 943: ditto

verifier/testreport.py
```diff
@@ -467,8 +467,11 @@ def create_html_report(self):
         # Flatten and combine the dictionary values
         fail_characterized = self.characterize_results_by_options(self.failing_tests, 'fail')
         fail_simple_diffs = self.check_simple_text_diffs(self.failing_tests, 'fail')
-        flat_combined_dict = self.flatten_and_combine(fail_characterized,
-                                                      fail_simple_diffs)
+        fail_list_differences = self.check_list_differences(self.failing_tests, 'fail')
```
echeran, line 470: Either put a condition here to guard this callsite only if the expected/actual is a list, or else take the list-specific handling logic of the new function and fold it into the existing function and just use that function instead.

verifier/testreport.py
```diff
@@ -921,6 +927,28 @@ def check_simple_text_diffs(self, test_list, category):
...
+            results[check] = set()
+
+        for fail in test_list:
+            label = fail['label']
+            actual = fail['result']
+            expected = fail.get('expected', None)
+            if type(actual) != type(expected):
+                # This is a type mismatch
+                continue
+            if type(actual) == list:
```
echeran, line 944: If this list-specific `check_list_differences` is kept, then this condition should go at the callsite, not here.

verifier/testreport.py
```diff
@@ -921,6 +927,28 @@ def check_simple_text_diffs(self, test_list, category):
...
+                if len(actual) != len(expected):
+                    results['different lengths'].add(label)
+                else:
+                    results['other list difference'].add(label)
```
echeran, line 948: idea(optional): maybe here, or maybe in the failure/error test characterization in the frontend, give an indication of the index number of the first element that diverges across the 2 compared lists, and also produce those element values.

echeran (commented):
Please include `segmenter.js` in this PR.

Before doing that, given that it's absent, the CI job should not be succeeding. Let's make it fail appropriately first, and push the commits to Github to retrigger CI. After that, then include the file in the PR and push.

schema/segmenter/test_schema.json
sven-oly: OK

schema/segmenter/result_schema.json
sven-oly: OK

testdriver/datasets.py
sven-oly: Done

verifier/testreport.py
sven-oly: Hmm, I can't test here because the functions check_simple_test_diffs and check_list_differences each operate on the list of failing tests. We probably need a check for each test result and the associated expected value.

verifier/testreport.py
sven-oly: See above.

verifier/testreport.py
sven-oly: Updating to note that results are not as expected due to a difference in data types. Then skip the string comparison characterizations.

.gitignore
```diff
@@ -21,6 +24,14 @@ DDT_DATA/testReports*/*
 
 TEMP_DATA
 
+# Generator and run config files. Ignore non-standard ones
+gen*.sh
+run*.json*
+run*.sh
+!generateDataAndRun.sh
+!run_config.json
+!genData100.sh
```
echeran, line 33: This seems heavy handed to ignore patterns of files yet allow-list specific exceptions for very important files. The way `.gitignore` files should be used is to list file patterns or specific files that we don't ever want to consider important.

If you have files that fit any of the patterns above that you are trying to ignore, then they are either:
1. temporary files that you should get rid of once they've served their purpose
2. important files that should be checked in

.gitignore
```diff
@@ -53,3 +64,6 @@ dependency-reduced-pom.xml
 
 # Dart
 executors/dart/.dart_tool/
+executors/dart_native
+executors/dart_web
```
echeran, line 68: If there are important source files in either `executors/dart_native` or `executors/dart_web`, then we shouldn't add the whole directory like this to the ignore file.

executors/node/segmenter.js
```diff
@@ -0,0 +1,57 @@
...
+
+module.exports = {
+  testSegmenter: function (json) {
+    const label = json['label'];
+    const locale = json['locale'];
+    let options;
+    let test_option;
+    if (json['options']) {
+      options = json['options'];
+      let desired_granularity;
```
echeran, line 14: suggestion: rename `desired_granularity` to either
* `ecma_granularity`
* `intl_granularity`
* `intl_api_granularity`
* `js_segmenter_granularity`

"desired" is not as clear and informative as it could be about how it's being used.

testgen/generators/segmenter_gen.js
```diff
@@ -0,0 +1,181 @@
...
+const common_fns = require("./common.js");
+const gen_hash = require("./generate_test_hash.js");
+
+const fs = require('node:fs');
+
+let debug = false;
+
+const test_type = "segmenter";
+
+const granularity = ['grapheme', 'word', 'sentence'];
```
echeran, line 20: add comment that Intl.JS doesn't support line breaking. include this link for reference to the reason:
https://github.com/tc39/proposal-intl-segmenter#why-is-line-breaking-not-included

verifier/testreport.py
sven-oly: Restructuring to report correctly.

testgen/generators/segmenter_gen.js
sven-oly: No, I'll include line break as an option, then mark as unsupported in the NodeJS executor.

executors/node/segmenter.js
sven-oly: Clarified with variable name ecma_intl_test_option.

.gitignore
sven-oly: Hmm, I have a set of gen scripts that I used for quick tests, e.g., testing only ICU75. But I don't think they are relevant to everyone.

verifier/testreport.py
sven-oly: This is a useful thought for looking at the differences for a particular test between expected and result. However, making a categorization for each will be too specific. This could be handled by a function for hovering over the result, similar to that now used for string differences.

verifier/testreport.py
sven-oly: restructuring this

verifier/testreport.py
sven-oly: Updating to test each specific failure for list differences.

.gitignore
echeran: Let's remove the remaining 3 lines (which are now the new lines 28-30) that are exceptions to the previous 3 lines (the previous lines 28-30) since the exceptions are now superfluous.

verifier/testreport.py
```diff
@@ -917,10 +927,39 @@ def check_simple_text_diffs(self, test_list, category):
...
+        #     results[check] = set()
+
+        label = test['label']
+        actual = test['result']
+        expected = test.get('expected', None)
+
+        if len(actual) != len(expected):
+            results['different lengths'].add(label)
+        else:
+            results['other list difference'].add(label)
```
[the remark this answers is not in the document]
sven-oly: Done

verifier/testreport.py
```diff
@@ -806,32 +809,39 @@ def add_to_results_by_key(self, label, results, input_data, test, key_list):
...
-    def check_simple_text_diffs(self, test_list, category):
+    def check_text_diffs(self, test_list, category):
         results = defaultdict(list)
         all_checks = ['insert', 'delete', 'insert_digit', 'insert_space', 'delete_digit',
                       'delete_space', 'replace_digit', 'replace_dff', 'replace_diff', 'whitespace_diff',
                       'replace', 'diff_in_()', 'parens', '() --> []', '[] --> ()',
-                      'comma_type', 'unexpected_comma']
-
+                      'comma_type', 'unexpected_comma', 'result_type_difference', 'boolean_difference',
+                      'error_in_key']
```
echeran, line 818: Should we add `other_list_differences` here, too?

echeran: I see that it's done, and you correctly put it in the singular, despite my comment above putting it in the plural.

echeran (approved):
LGTM

PR #460, opened 2025-05-22
Disable Dart Native in order to unblock CI

echeran:
Current PRs with changes unrelated to Dart are starting to [fail all of a sudden](https://github.com/unicode-org/conformance/actions/runs/15171956221/job/42738320686). This wasn't a problem in the [latest run on `main`](https://github.com/unicode-org/conformance/actions/runs/14935490521/job/41961798188) 2 weeks ago

### DIFF
--- a/.github/workflows/dart.yaml
+++ b/.github/workflows/dart.yaml
@@ -47,12 +47,15 @@ jobs:
         if: always()
 
       - run: dart test
-        if: ${{ matrix.run-tests }}
+        if: false  # reenable when Dart Native is not causing errors (fixed/removed)
+        # if: ${{ matrix.run-tests }}
 
       - run: dart test -p chrome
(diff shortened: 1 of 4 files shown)


sven-oly (approved):
Thank you!

echeran:
The point of this PR is to do the minimal work to unblock ongoing development work in other PRs like #461. Initially, the thought was just to disable the flags that enabled the experimental features, based on the [log output from the blocked PRs](https://github.com/unicode-org/conformance/actions/runs/15171956221/job/42738320686#step:7:674):
```
Unavailable experiment: native-assets (this experiment is only available on the main, dev channels, this current channel is stable)
Unavailable experiment: record-use (this experiment is only available on the main, dev channels, this current channel is stable)
```
However, the build command itself depends on the experimental features being enabled. This seems related to native compilation. So the current thought is to disable `dart_native` while still keeping enabled the `dart_web` platform.

PR #465, opened 2025-06-11
ICU4X Segmenter now working for 2.0 beta2

sven-oly:
Note that there are differences between ICU4X results and the other platforms.


sven-oly:
Note that 2 test failures are found in Myanmar text for word and line breaks.

Issue #471, opened 2025-06-30
ICU4J collation: old-style collation does not work. But it does work in ICU4C!


sven-oly:
This is data in ICU76.1 that includes an old style of collation locale de@coLLation=PhoneBook". This test fails in ICU4J but passes in the CPP version. Changing the casing of the locale does not fix it.

This test passes when the locale is set this way: `"locale":"de-u-co-phonebk"`

```
{
  "test_type": "collation",
  "compare_type": "&lt;2",
  "s1": "ae",
  "s2": "ä",
  "source_file": "collationtest.txt",
  "line": 2426,
  "label": "01362",
  "locale": "de@coLLation=PhoneBook",
  "test_description": "locale @collation=type should be case-insensitive",
  "hexhash": "893bdab906f7bc0e918ce388c7e78799d5913eaa"
}
```

sffc:
It seems that the indicated test `01362` is passing.

sven-oly:
This has been resolved by updating test generation.

PR #472, opened 2025-07-03
Collation data generation fixes, reducing test failures

sven-oly:
#471 

This should change the number of test failures in ICU76 as follows:
ICU4C:  1557 --> 1444
ICU4J:     200 --> 163
ICU4X:    389 --> 319

More fixes to come!

### DIFF
--- a/executors/node/collator.js
+++ b/executors/node/collator.js
@@ -8,8 +8,10 @@ module.exports = {
     let outputLine = {'label':json['label']};
 
     // Locale if provided in the test data.
-    let testLocale = undefined;
-    testLocale = json['locale'];
+    let testLocale = 'en';  // default
+    if ('locale' in json) {
+      testLocale = json['locale'];
+    }
 
     if (testLocale == 'root') {
       outputLine =  {'label': json['label'],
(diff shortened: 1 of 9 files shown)


sven-oly:
Also, overall Dart failures are also reduced, but there are new test errors with this. "Collator compare failed" is the new error.

sven-oly:
NodeJS failures are gone, but there are new unsupported reports including many "RangeError" problems. This is now checking for supported locales, which is causing the increase in "unsupported".

sven-oly:
I found a little problem with test generation when the string compared has white space, e.g., "REE, B". The last part was being deleted. This fixes 2 to 5 tests in the executors.

executors/node/collator.js
```diff
@@ -30,11 +69,10 @@ module.exports = {
         testCollOptions['sensitivity'] = 'accent';
       } else
       if (strength == 'tertiary') {
-        testCollOptions['sensitivity'] = 'case';
+        testCollOptions['sensitivity'] = 'variant';
```
sffc, line 72: Hmm, how many cases does this fix?

sffc (approved):

executors/node/collator.js
sven-oly: I'm not sure but will try to figure this out.

sven-oly:
A problem found with CPP and collation: the JSON-C library uses c-string forms, which means that any string with \u000 is terminated at that point. This accounts for some test failures.

PR #475, opened 2025-07-11
Use workaround to fix many collation failures in ICU4C.

sven-oly:
Test generation in Python puts SMP characters (> 0xffff) as \ud8** surrogate pairs. This works OK for ICU4J, ICU4X, and NodeJS, but fails with the current CPP JSON.

This workaround adds two new items that contain the hex codes for the characters to be compared in s1 and s2. These are set for the NON_IGNORABLE and SHIFTED data files.

This is admittedly a hack, but it points out that the tests with SMP characters were not working. With this hack, over 1200 tests now pass that were previously failing.

### DIFF
--- a/executors/cpp/coll.cpp
+++ b/executors/cpp/coll.cpp
@@ -42,10 +42,60 @@ using icu::RuleBasedCollator;
 
 const char error_message[] = "error";
 
+UnicodeString get_char_from_hex_list(json_object* str_codes_obj,
+                                     int debug_level) {
+    // Get the hex codes and assemble into a string with \u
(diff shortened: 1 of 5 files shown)


sven-oly:
I want to submit this, noting the hack as an issue to be resolved. WDYT?

sffc (approved):
I have mixed feelings about this PR; it's nice that it gets C++ passing more test cases, but it is an inelegant solution that impacts the test data for all implementations.

I will approve it, but please open a follow-up issue to clean up the tech debt.

PR #476, opened 2025-07-11
Fix almost all NodeJS collation issues.

sven-oly:
This acknowledges that -kr- is not supported as a locale extension,

And -co-search is supported as a setting for the "usage" option.


echeran (approved):
LGTM

PR #478, opened 2025-07-15
Collation node - fix last test failures

sven-oly:


sven-oly:
This PR moves the last test failures to either "unsupported" or "pass". One open question is whether "root" locale should be replaced with something else instead of being marked as unsupported.

PTAL

echeran (approved):

PR #484, opened 2025-07-17
Start to sort columns of matrix by count of components

sven-oly:


sven-oly:
Not needed because #502 superseded this PR.

Issue #489, opened 2025-07-23
Use large collation tests only for new releases of executors


sven-oly:
Idea: don’t tun collation tests on the 400K simple tests. All of them pass in C,J,X, and NodeJS when the executors get the right data, and they don’t change much.

Suggestion: add a flag to the test generator that does create these tests, but otherwise doesn’t. Turn this on only for each new executor release. i.e., ICU79

sven-oly:
Another idea: make the exhaustive test a separate component, only for the latest ICU release.

Or sample the big tests, perhaps 1%?

PR #490, opened 2025-07-23
Make large collation tests optional with command line

sven-oly:
This removes the very large collation sets from test generation *unless* the generate script adds the option explicitly as a command line argument.

This shortens the test generation time and also reduces the size of the data.

#489


sven-oly:
This is not necessary now. We have found a problem with JSON-C that causes > 1200 test failures when SMP characters are compared. PR #493 will mark those as "known issues".

Issue #467, opened 2025-06-16
Re-enable the Dart Native executor


sffc:
It was disabled in https://github.com/unicode-org/conformance/pull/460 because it was breaking CI. There are stability issues since this depends on unstable Dart toolchains. We should re-enable the executor in CI, but do it in a way that won't break as often. Is there a way to pin the specific nightly version, for example? @echeran says that the old toolchain installation doesn't stay in sync with the ICU4X dependency. This smells like a solvable problem, but we should solve it in a clean way.

@mosuem

mosuem:
We could pin both the Dart and the `package:intl4x` version - that way it definitely won't break. To update, I will periodically file PRs, until we are stable enough so we can unpin.

PR #492, opened 2025-07-25
Add Dart native

mosuem:
Remove `likely_subtags` as this is not supported in `Intl4x` at the moment, but add `plural_rules` and `list_format`.


sven-oly:
This would be a very useful update. Please advise how to updated to Dart 3.8.0 so I can run this. We'll also have to integrate that into the end-to-end workflow.

.github/workflows/e2e.yml
```diff
@@ -32,6 +32,8 @@ jobs:
 
       - name: Setup version of Dart
         uses: dart-lang/setup-dart@d6a63dab3335f427404425de0fbfed4686d93c4f
+        with:
```
echeran, line 35: This reintroduces a dev version (non-stable) of the Dart runtime...

echeran (changes requested):
@mosuem We have issue #467 filed to describe the work you're tackling to reenable the Dart Native executor. It references PR #460, in which we had to disable the Dart Native executor because the Conformance project is about testing stable libraries on stable toolchains. Please ensure that you will uphold that invariant in this PR, if that isn't already your plan as you work on this PR. This principle will help us avoid the previous problems that became an obstacle to merging PRs that required #460 and #467 in the first place.

.github/workflows/e2e.yml
mosuem: Pinned it!

executors/dart/pubspec.yaml
```diff
@@ -2,17 +2,17 @@ name: dart_executor
...
-  sdk: ^3.7.0
+  sdk: ^3.10.0-37.0.dev
 
 # Add regular dependencies here.
 dependencies:
   collection: ^1.19.1
-  intl4x: ^0.11.3
+  intl4x: ^0.12.2
   node_preamble: ^2.0.2
   pubspec_lock_parse: ^2.2.0
```
sffc, line 12: @mosuem Can you pin these versions (not with `^`) to reduce the risk of breakages? (for `intl4x` and `sdk`)'

Or, can we guarantee that we always install form the lockfile?

sffc (approved):

sven-oly:
Trying to merge now.

PR #495, opened 2025-08-05
Fix collation test gen. Also resolve all Java and CPP executor test failures.

sven-oly:
This resolves all ICU4J collation tests.

It also reduces ICU4C test failures from 79 to 7.

### DIFF
--- a/executors/cpp/coll.cpp
+++ b/executors/cpp/coll.cpp
@@ -272,6 +272,42 @@ auto TestCollator(json_object *json_in) -> string {
       }
     }
 
+    json_object *alternate_obj = json_object_object_get(json_in, "alternate");
+    if (alternate_obj) {
+      // TODO: Check status
+      string alternate = json_object_get_string(alternate_obj);
+      if (alternate == "shifted") {
+        rb_coll->setAttribute(UCOL_ALTERNATE_HANDLING, UCOL_SHIFTED, status);
+      } else
+      if (alternate == "non-ignorable") {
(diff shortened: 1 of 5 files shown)


sven-oly:
Fixing a problem in test generation resolved almost all of the problems. Then processing the options for caseFirst and caseLevel resolved the last of the ICU4J failures:
image

This also fixes almost all ICU4C test failures (investigating the last 6 now!)

Please take a look!

executors/cpp/coll.cpp
```diff
@@ -292,12 +346,28 @@ auto TestCollator(json_object *json_in) -> string {
...
+    if (alternate_obj) {
+      uni_coll->setAttribute(UCOL_ALTERNATE_HANDLING, alternate_value, status);
+      if (check_icu_error(status, return_json, "alternate")) {
+        json_object_object_add(
+            return_json, "actual_options",
+            json_object_new_string(json_object_get_string(json_in)));
+
+        return json_object_to_json_string(return_json);
       }
+    }
```
echeran, line 368: What's the difference between this and lines 290-299?

sffc: It appears that the first instance is gated by `if (rules_len > 0)` which uses a different code path to configure the runtime-rules collator. It would probably clean up the code to do more code sharing between these two code paths, but it is a pre-existing issue.

sven-oly: Yes, that's right, Shane. This needs to be refactored. I'll add an issue.

PR #496, opened 2025-08-06
Fix getting data from output tables with left click with HTML escape/…

sven-oly:
…unescape. This resolves incomplete data being shown in test output areas because of "&", "<", etc.

Note that this uses simple replace to handle the problems expected in collation test data.

### DIFF
--- a/verifier/detail_template.html
+++ b/verifier/detail_template.html
@@ -388,7 +388,7 @@ table th {
                         } else {
                             output = item[key];
                         }
-                        html.push('<td onclick=' + onclick_call +'>' + output +'</td>');
+                        html.push('<td onclick=' + onclick_call +'>' + escapeHtml(output) +'</td>');
                     }
                 }
                 html.push("</tr>");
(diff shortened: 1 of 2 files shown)


verifier/detail_template.html
```diff
@@ -615,7 +615,20 @@
...
+        .replace(/&/g, "&amp;")
+        .replace(/</g, "&lt;")
+        .replace(/>/g, "&gt;")
+        .replace(/"/g, "&quot;")
+        .replace(/'/g, "&#039;");
+      }
+      return unsafe;
+   }
+
+function clearSelectedItems(the_button, test_class) {
```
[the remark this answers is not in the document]
sven-oly: OK.

verifier/detail_template.html
```diff
@@ -388,7 +388,7 @@
                         } else {
                             output = item[key];
                         }
-                        html.push('<td onclick=' + onclick_call +'>' + output +'</td>');
+                        html.push('<td onclick=' + onclick_call +'>' + escapeHtml(output) +'</td>');
```
sffc, line 391: Thought: We should ideally be building this up using a proper templating engine or else as a proper DOM tree so that we can write to `innerText` properties and the escaping is done automatically for us; it is bad practice to build up HTML as a string. But, that's cleanup we should do later, outside of the scope of this PR.

verifier/detail_template.html
```diff
@@ -615,30 +615,50 @@
...
+      if (typeof unsafe === 'string') {
+        return unsafe
+        .replace(/&/g, "&amp;")
+        .replace(/</g, "&lt;")
+        .replace(/>/g, "&gt;")
+        .replace(/"/g, "&quot;")
+        .replace(/'/g, "&#039;");
+      }
+      return unsafe;
+    }
```
sffc, line 629: Thought: I don't see a build-in JS function for this. Seems you either write this function yourself or use the DOM. I think most people don't need this function any more because they use a templating engine.

verifier/detail_template.html
```diff
@@ -615,30 +615,50 @@
...
-            // Set the current count of items
-            output[0].innerHTML = output[0].innerText = newSize;
+          // Set the current count of items
+          output[0].innerHTML = output[0].innerText = newSize;
         }
     }
 
-    function unEscape(htmlStr) {
-        var doc = new DOMParser().parseFromString(htmlStr, "text/html");
-        return doc.documentElement.textContent;
```
sffc, line 641: Question: Why did you change the implementation of this function? It seems correct. You can even do the reverse operation to perform the escaping.

PR #507, opened 2025-08-13
Remove unescaping from test strings, fixing collation test failures.

sven-oly:


sven-oly (commented):
This fixes the problems, making collation cleaner.

PR #509, opened 2025-08-18
[Dart] Fix some issues and upgrade to new DateTime API

mosuem:
This builds on a newer intl4x DateTime API and adresses some of the points in https://github.com/dart-lang/i18n/issues/1000 and https://github.com/dart-lang/i18n/issues/998

### DIFF
--- a/executors/dart/test/misc_test.dart
+++ b/executors/dart/test/misc_test.dart
@@ -2,8 +2,11 @@ import 'dart:async';
 import 'dart:convert';
 
 import 'package:dart_executor/collator.dart';
+import 'package:dart_executor/datetime_format.dart';
 import 'package:dart_executor/lang_names.dart';
 import 'package:dart_executor/numberformat.dart';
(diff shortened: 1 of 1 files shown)


sven-oly:
Hey Moritz, Thank you for working on this. There are some test failures here. Please try to resolve these so we can integrate.

mosuem:
Yes, this is waiting for https://github.com/dart-lang/i18n/pull/1003 to land. But even then, there will be still quite a lot of failures in the datetime formatting. Better than before though!

mosuem:
> `Unknown combination of`

This sounds like a JS error, so there must be some wrong datetime option parsing on my part.

> `UNKNOWN`

I always return the original label, so it must already have been UNKNOWN. Do you have an example?

sven-oly (approved):
This looks OK as far as I can tell.

mosuem:
Looking good!

PR #510, opened 2025-08-19
Datetime update options

sven-oly:
Fixes over half of the test failures in NodeJS datetime_fmt for ICU77.


sffc (commented):
Seems about right!

sven-oly:
PTAL. Resolved all the comments and suggestions.

PR #511, opened 2025-08-19
Datetime format tests: remove before 76

sven-oly:


sven-oly:
Removed the generator for datetime_fmt tests that used NodeJS. That was only used before ICU76.

sffc (approved):
This looks like more what I expected to see

PR #522, opened 2025-09-02
Rationalize references to "lang_names" in Python routines

sven-oly:


sven-oly:
Removed the old data generator.

echeran (approved):

Issue #459, opened 2025-05-21
End-to-end should quit for Node JS when executor missing


echeran:
At the time of leaving [this comment](https://github.com/unicode-org/conformance/pull/458#pullrequestreview-2844963121), the PR was passing CI checks even though the executor code in question for Node JS had not yet been added to the PR.

We should make sure that if the executor code doesn't exist, then either the executor code should report all errors, or else the pipeline should catastrophically quit early.

I want to use the word "quit" or "exit" to indicate behavior that our pipeline should take as a last resort, so that we make noise loudly and early. We should avoid succeeding silently, as we currently are, because it means we never know when we have problems.

PR #528, opened 2025-09-08
Handle missing executor in NodeJS

sven-oly:
#459


schema/collation/test_schema.json
```diff
@@ -111,7 +111,11 @@
           "warning": {
             "description": "Present when data has a problem",
             "type": "string"
-          }
+          },
+          "hash_id": {
+            "description": "Has value of the JSON string omitting label value",
+            "type": "integer"
+         }
```
sffc, line 118: Question: I don't see `hash_id` used anywhere in this PR outside the schema. Why do you need it?

PR #531, opened 2025-09-09
Update logging with info on the source file running

sven-oly:
Adds the name of the python file being executed in logging output. 

Also removes old unneeded code.

Add npm command to handle package-lock.


sffc (commented):
This is good progress. I'm eager to see the end-to-end log when CI is passing. It seems you are annotating a lot of the messages in Python, but I think you might also want to annotated the messages generated by the bash scripts as well as any generated by a Python subprocess.

generateDataAndRun.sh
```diff
@@ -22,6 +22,8 @@ fi
 # Install NVM if it is not install in CI
 
 export NVM_DIR=$HOME/.nvm
+# To clear lock
+npm install --package-lock-only
```
sffc, line 26: Same

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "@js-temporal/polyfill": "^0.5.1"
+  }
+}
```
sffc, line 5: We should have a package.json in each node package directory, but probably not at the top level

PR #532, opened 2025-10-08
Fix platform order

sven-oly:
That was somehow broken, but this makes it work.


sffc (approved):
I don't understand this part of the code but this looks plausible

sven-oly:
Revisiting this PR, hoping to integrate.

PR #543, opened 2026-04-27
Fixes to make genData100 work on Mac

sven-oly:
Created in collaboration with @sffc


echeran (approved):
hmm, changes look good, but not sure why CI fails. that should be figured out before merging.

sffc:
There are lots of errors so I'm not sure which errors are most important. I'll see if they reproduce on main: https://github.com/unicode-org/conformance/pull/544

sffc:
Seems like it succeeds some fraction of the time. Not clear what causes it. I will keep kicking it until it succeeds

PR #544, opened 2026-04-28
Switch CI to use a stable container

sffc:
No-op change to test whether CI is clean

### DIFF
--- a/schema/schema_validator.py
+++ b/schema/schema_validator.py
@@ -26,7 +26,7 @@ ch.setLevel(logging.WARNING)
 
 
 def parallel_validate_schema(validator, file_names):
-    num_processors = mp.cpu_count()
+    num_processors = min(2, mp.cpu_count())
     logging.info('JSON validation: %s processors for %s plans', num_processors, len(file_names))
 
     # How to get all the results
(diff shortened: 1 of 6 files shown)


echeran:
Ah, this CL proves that `main` is not happy at the moment.

sffc:
CI passed 4 months ago but the logs are not retained.

There are so many errors I don't know where to start.

I wonder if it is a Python version error because there are a lot of what seem like Python errors, things like

```
  File "/usr/lib/python3.12/logging/__init__.py", line 392, in getMessage
    msg = msg % self.args
          ~~~~^~~~~~~~~~~
TypeError: not all arguments converted during string formatting
```

And most of the errors later on seem to be "cannot find testdata", which could occur if the testdata generation script was failing due to python version mismatch.

echeran (approved):
LGTM

schema/schema_validator.py
```diff
@@ -26,7 +26,7 @@
 
 
 def parallel_validate_schema(validator, file_names):
-    num_processors = mp.cpu_count()
+    num_processors = min(2, mp.cpu_count())
```
sffc, line 29: Note: this was a bit of a hack, but it seemed like the CI was regularly crashing in the validater step, so Gemini decided to try to limit the parallelism. Ideally this would be a flag or environment variable.

sffc:
Note: the "create disk space" steps no longer work since the `ubuntu:24.04` container cannot access the files inside the host image. However, it seems GitHub significantly increased the disk space available to this job; it was around 20 GB before and now it is 80 GB. So, I'm inclined to land this, because I value the stability of the operating system image.

